PR #12, opened 2021-10-21
fix #11

### LINKED ISSUE #11: Can't parse simple command
Hi,
I'm new to IMAP protocol, so I may be wrong.

When I run:
* cargo run --example=parse_command
* Enter IMAP4REV1 command (or "exit"): a3 FETCH 1 RFC822
I get:
* Error parsing the command. Is it correct? (Error(Error { input: [70, 69, 84, 67, 72, 32, 49, 32, 82, 70, 67, 56, 50, 50, 10], code: Tag }))

If I understand correctly BODY[] == RFC822.
Trying "a3 FETCH 1 BODY[]" works as intended.
Should I replace all plain RFC822 to BODY[] or is this some bug in the library?

FelixPodint:
Seems to easy to be correct. Variant was already in place, so was serialization.
Tested with example and my own project and everything seems to work.
Is it ok for imap_codec::types::response::DataItemResponse::Rfc822(NString) to have NString or should it be imap_codec::types::body::BodyStructure like imap_codec::types::response::DataItemResponse::Body(BodyStructure) is?

### DIFF
--- a/src/parse/command.rs
+++ b/src/parse/command.rs
@@ -495,6 +495,7 @@ fn fetch_att(input: &[u8]) -> IResult<&[u8], DataItem> {
         value(DataItem::Rfc822Header, tag_no_case(b"RFC822.HEADER")),
         value(DataItem::Rfc822Size, tag_no_case(b"RFC822.SIZE")),
         value(DataItem::Rfc822Text, tag_no_case(b"RFC822.TEXT")),
+        value(DataItem::Rfc822, tag_no_case(b"RFC822")),
     ))(input)
 }
 


duesee:
Could you move the new parser to be evaluated as the last one of the RFC822 variants, commit with `--amend`, and force push?

From my understanding we always want the longest matches to be evaluated first. Otherwise parsing `RFC822.SIZE` may result in the parsing of `RFC822` as `DataItem::Rfc822` with `.SIZE` as a remainder. I hoped that the tests would catch it. I will add some more tests later and see if that is the case.

Issue #15, opened 2022-03-18
Parsing Envelope


FelixPodint:
Hi,
Is there a way to parse raw bytes `&[8u]`, `Vec<u8>` to `Envelope` struct?

Thanks.
With respect,
Felix P.

duesee:
The [envelope](https://github.com/duesee/imap-codec/blob/6bf1e5d0da45d576bd9ed4ddc0b3640da8e2ba80/src/parse/envelope.rs#L31) function should be what you need. However, it is pub(crate), i.e., not exposed via a public API. Can you tell me a bit about your use case? Trying to understand if it would make sense to expose something like `use imap_codec::nom::envelope`.

duesee:
Actually, I like the idea of having an "ecosystem" of composable nom functions. Thus, I would be happy to expose (all) parsers via `imap_codec::nom::*`. Do you like to make that PR? :-) I will be at home in some hours and can review the PR (or commit something myself).

FelixPodint:
I'm using SQLite to store mail messages for my IMAP server. Currently I store envelope information as data blob inside this database. I don't know, maybe I should store all envelope fields as separate columns. That way I would just construct new `Envelope` struct from separate SLQ columns.

Please commit yourself, if its not a problem for you, I'll be out for a weekend.
Thanks.

FelixPodint:
Ok, before I do anything, I want to ask.
Do you want new `pub mod nom` with re-exports of functions from `mod parse` or do you want to replace `mod parse` with `mod nom`?
Creating a separate `mod nom` would allow us to aggregate all [from bytes to struct] functions into one place, while keeping `mod parse` structure intact. Not sure how to nicely put [from struct to bytes] function (aka. `.encode(writer)`) to `mod nom`.

FelixPodint:
I think `Decode` trait would be the best option. (Easiest to understand for new users)

duesee:
Hey @FelixPodint,

Sorry, I didn't had too much time to look into it. Thanks for taking care!

> Do you want new pub mod nom with re-exports of functions from mod parse or do you want to replace mod parse with mod nom?

I think mod parse should stay for now.

> Creating a separate mod nom would allow us to aggregate all [from bytes to struct] functions into one place, while keeping mod parse structure intact.

[...]

PR #16, opened 2022-03-23
Exposed a bunch of parse functions to the public API

FelixPodint:
For issue #15 
Exposed some parsing functions from `mod parse`.
Left `crate::parse::core` untouched. Not sure what is useful in `core` for most users.
Also couldn't export some functions that used non-public types, such as `AStringRef`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2018"
 [features]
 default = []
 serdex  = ["serde", "chrono/serde"]
+nomx = []
 
 [dependencies]
 arbitrary = { version = "1", optional = true, features = ["derive"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,6 +6,8 @@ use codec::Encode;
 pub mod arbitrary;
 pub mod codec;
 pub mod parse;
+#[cfg(feature = "nomx")]
+pub mod rfc3501;
 pub mod state;
 pub mod types;
 pub mod utils;
--- a/src/parse/address.rs
+++ b/src/parse/address.rs
@@ -14,7 +14,7 @@ use crate::{
 ///               addr-adl SP
 ///               addr-mailbox SP
 ///               addr-host ")"
-pub(crate) fn address(input: &[u8]) -> IResult<&[u8], Address> {
+pub fn address(input: &[u8]) -> IResult<&[u8], Address> {
     let mut parser = delimited(
         tag(b"("),
         tuple((addr_name, SP, addr_adl, SP, addr_mailbox, SP, addr_host)),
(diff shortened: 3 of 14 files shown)


duesee:
Thanks! I will have to learn what `#[doc(inline)]` does and try to check if everything is exported.

In the meantime, could you run `cargo fmt --config imports_granularity=Crate,group_imports=StdExternalCrate`, `git --amend`, and force push?

I will try to review it today. Have you tried to use this version in your email server, i.e., in `Cargo.toml` `imap_codec = { path = "..." }`, and see if you like it that way?

I am asking, because I thought about exposing all parsers directly on the first level of `internal::*`. The reason is that people will likely read the IMAP RFC and wonder if they could use a specific rule. However, this will make `parse` and `internal` go out of sync. Not sure what is the best solution.

FelixPodint:
> could you run cargo fmt --config imports_granularity=Crate,group_imports=StdExternalCrate

This command didn't work, but after changing to nightly and creating `rustfmt.toml` file simple cargo fmt worked. Is it ok?

> Have you tried to use this version in your email server

Yes, currently using local clone.

> and see if you like it that way?

Don't know yet. Everything still feels like a mess (on my end :P), will see...
```
// Little example:
FetchAttribute::Envelope => {
    let envelope_bytes = self.db.envelope(
        &self.user,
        mailbox,
        seq_or_uid.get() as usize,
        uid,
    )?;
    let envelope = match imap_codec::internal::envelope(&envelope_bytes) {
        Ok((remaining, envelope)) => {
            if remaining.is_empty() {
                envelope
            } else {
                return Err(anyhow!(
                    "Some bytes remained after parsing Envelope"
                ));
            }
        }
        Err(e) => {
            return Err(anyhow!(
                "Command parsing error: {}",
                e.map(|e| e.code)
            ));
        }
    };
    rez.push(FetchAttributeValue::Envelope(envelope));
}
```
> I am wondering if we should expose all parsers directly on the first level of `internal::*`.

No, there should be some structure to them.

>  However, this will make `parse` and `internal` go out of sync. Not sure what is the best solution.

Agree, this is why I was thinking about `Decode` trait. Implement it for all types, then appropriate, and don't expose `parser` to public API.

FelixPodint:
Done. For now `rfc3501` will do. It feels like a temporary solution for me. We'll discus [more](https://github.com/duesee/imap-codec/discussions/18).

duesee:
I would like to first squash your commits, then split them again, such that adding `rustfmt.toml` is the first separate commit and the rest the second, and then merge. I don't know how experienced you are with git . Do you prefer to make it yourself or can I do it? (Don't want to take away a learning experience from you :-))

duesee:
Argh, sorry. I had an outdated version...

duesee:
Hey, I merged your commit with a slightly different commit message (f56a20eaaa04720a8a9511fd31f3b127513a53df). Thanks, again! I also worked on the API. You can compile with `cargo doc --all-features` and see what I mean. IMAP extensions were refactored into a new module `extensions` and feature gated with `ext_*` (see `Cargo.toml`). This makes a bit easier to find them. Also, all nom parsers can now be accessed via `internal` and the `nomx` feature. Here, I also unified the documentation a bit.

PR #35, opened 2022-05-24
fix benchmarks

franziskuskiefer:


duesee:
Thanks, @franziskuskiefer! This code somehow got old and wasn't maintained. Probably there should also be an action that checks if things work. Or compile at least. I created issue #36 for now.

PR #250, opened 2023-06-06
Add annotations to feature-gated items

### LINKED ISSUE #232: Describe feature-gated methods/variants/etc. in documentation.
https://users.rust-lang.org/t/how-to-test-a-package-metadata-docs-rs-setting/53818, thanks, @jakoschiko.

jakoschiko:
This PR:

- Customizes the docs.rs build. When building the documentation docs.rs will now use `--all-features` and set the configuration flag `docsrs`. See [here](https://docs.rs/about/metadata) for reference.
-  Enables the nightly feature [doc_cfg](https://doc.rust-lang.org/unstable-book/language-features/doc-cfg.html) if the configuration flag "docsrs" is set. It allows us to add annotations to feature-gates items. For this it's necessary to build the documentation with a nightly compiler.
- Adds annotations to all feature-gates items for the features `tokio`, `unchecked`, `starttls` and `ext_*` via `doc(cfg(feature = "..."))`.

There are some caveats:

- Even though it's possible to add an annotation to a feature-gated trait implementation (`impl Foo for Bar`), that's not possible for derived trait implementations (`#[derive(Foo)] struct Bar`). This effects `serde`, `arbitrary` and `bounded-static`. I decided to not annotate any trait implementations for these libraries, even if the traits are implemented manually for some types. I suggest to document these feature-gates on the top-level page.
- It's not possible to add an annotation to a feature-gated function parameter (e.g. the parameter `sync` of the function `Literal::unchecked`). I suggest to document the feature-gate in the function documentation.

The result can be tested with:
```
RUSTFLAGS="--cfg docsrs" RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --no-deps --all-features
```

The result looks like this:
image

Resolves #232

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -356,6 +356,7 @@ pub fn authenticate(input: &[u8]) -> IResult<&[u8], AuthMechanism> {
 ///                CRLF is parsed by upper command parser.
 /// ```
 #[cfg(feature = "ext_sasl_ir")]
+#[cfg_attr(docsrs, doc(cfg(feature = "ext_sasl_ir")))]
 #[allow(clippy::type_complexity)]
 pub fn authenticate_sasl_ir(
     input: &[u8],
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,10 @@ harness = false
 name = "parse_response"
 harness = false
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
 [workspace]
 members = [
     "fuzz",
--- a/imap-types/Cargo.toml
+++ b/imap-types/Cargo.toml
@@ -44,3 +44,7 @@ thiserror = "1.0.29"
 [dev-dependencies]
 criterion = "0.4"
 rand = { version = "0.8", default-features = false, features = ["small_rng"] }
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/imap-types/src/command.rs
+++ b/imap-types/src/command.rs
@@ -156,6 +156,7 @@ pub enum CommandBody<'a> {
     /// STARTTLS.  The server MAY advertise different capabilities after
     /// STARTTLS.
     #[cfg(feature = "starttls")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "starttls")))]
     StartTLS,
 
     /// ### 6.2.2.  AUTHENTICATE Command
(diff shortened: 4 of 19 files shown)


duesee:
Awesome, thanks! 🎉 I took the liberty to add `RUSTFLAGS="--cfg docsrs"` to your description as you helpfully pointed out yesterday :-) Works like a charm!

Regarding your first caveat: I think this is totally fine for the crates you mentioned. I'll file an issue to describe these features separately.

Regarding your second caveat:

I think the caveats you mention may hint to a few issues in imap-codec itself. Notably, the way in which some feature-gates are used could unnecessarily break existing code. For example, there is no reason someone has to change `CommandBody::authenticate(...)` to `CommandBody::authenticate(..., None)` just because they activated the `ext_sasl_ir` feature. We could provide a new method `CommandBody::authenticate_with_ir` instead and free two birds with one key.

My suggestion would be to merge this PR as is and file some follow-ups. What do you think?

Follow-ups:

* [ ] Improve `serde`, `arbitrary`, and `bounded-static` feature documentation.
* [ ] Add feature-gated `CommandBody::authenticate_with_ir` constructor.
* [ ] Add a feature-gated `Literal::unchecked_non_sync` constructor.
* [ ] Privatize `Literal::sync` and add a feature-gated `Literal::is_sync` method.

(There might be some more work items.)

duesee:
We could also add `#[cfg_attr(docsrs, doc(cfg(feature = "ext_*")))]` to `mod extensions;`

src/command.rs
```diff
@@ -333,6 +333,7 @@ pub fn password(input: &[u8]) -> IResult<&[u8], AString> {
 ///                CRLF is parsed by upper command parser.
 /// ```
 #[cfg(not(feature = "ext_sasl_ir"))]
+#[cfg_attr(docsrs, doc(cfg(feature = "ext_sasl_ir")))]
```
duesee, line 336: There should be a `not` here. But as discussed, let's not try to do this now and resolve it properly in a follow-up.

jakoschiko: I would remove this in this PR because this annotation is simply wrong.

jakoschiko:
> I took the liberty to add `RUSTFLAGS="--cfg docsrs"` to your description as you helpfully pointed out yesterday :-) Works like a charm!

Last time I checked it worked with only `RUSTDOCFLAGS="--cfg docsrs"` (but not with only `RUSTFLAGS="--cfg docsrs"`).

> We could also add #[cfg_attr(docsrs, doc(cfg(feature = "ext_*")))] to mod extensions;

I'll add this in this PR.

I'm okay with fixing the remaining issues later.

duesee:
Couly you also run `cargo +nightly fmt` to make the CI happy? I guess you are fine when we squash this?

jakoschiko:
> Could you also run `cargo +nightly fmt` to make CI happy? I guess you are fine squashing this?

Done

PR #282, opened 2023-06-17
chore: Set imap-codec version to 0.9.1.

duesee:


duesee:
Thanks, cargo-semver-checks!

PR #330, opened 2023-08-02
Resolve (some) SemVer issues.

duesee:


duesee:
Let's get this in. It's not complete but we reduced two obvious hazards by making `Capability` and `CapabilityEnable` `#[non_exhaustive]`.

Issue #332, opened 2023-08-02
misuse: Exclude non-sync literals in responses.


duesee:
> The non-synchronizing literal form MUST NOT be sent from server to client.

-- https://datatracker.ietf.org/doc/html/rfc7888

Currently, it can.

PR #333, opened 2023-08-03
fix: Rework `DecodeError`.

duesee:
I would argue the `tag` is required in `LiteralFound`, so doing this breaking change is fine.

The implementation is... peculiar. But we are limited by the current parsing architecture. The only way we can "signal" that a literal was found is through `IMAPParseError { kind: IMAPErrorKind::Literal { .. } }`. This structure can be returned in any of the deeply-nested `fn literal` parsers that don't have access to anything before, i.e., the tag.

Edit: I have a better idea: The `fn literal` parser now trusts that the upper `fn command` parser will fill in the missing `tag`. This way, we don't need to parse the `tag` twice.

What do you think, @jakoschiko?

### DIFF
--- a/imap-codec/src/codec/decode.rs
+++ b/imap-codec/src/codec/decode.rs
@@ -1,7 +1,7 @@
 use std::num::{ParseIntError, TryFromIntError};
 
 #[cfg(feature = "bounded-static")]
-use bounded_static::IntoBoundedStatic;
+use bounded_static::{IntoBoundedStatic, ToStatic};
 #[cfg(feature = "ext_literal")]
 use imap_types::core::LiteralMode;
 #[cfg(feature = "ext_idle")]
@@ -9,7 +9,8 @@ use imap_types::extensions::idle::IdleDone;
 use imap_types::{
     auth::AuthenticateData,
     command::Command,
-    response::{Continue, Greeting, Response},
+    core::Tag,
+    response::{Greeting, Response},
 };
 use nom::error::{ErrorKind, FromExternalError, ParseError};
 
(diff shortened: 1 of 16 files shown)


jakoschiko:
I don't have any opinion about the implementation, but a `Tag` inside of `LiteralFound` is fine for me. Currently it's a `Option<Tag>`. Does this reflect the IMAP standard? Is it possible to construct a `Command` with literals but without `Tag`? If yes, can the server reject this `Command`?

duesee:
> Currently it's a `Option<Tag>`. Does this reflect the IMAP standard? Is it possible to construct a `Command` with literals but without `Tag`?

No, it doesn't reflect the IMAP standard :-/ Every command in IMAP4rev1 must have a tag. ABNF:

```abnf
command = tag SP (command-any / command-auth / command-nonauth / command-select) CRLF
```

> If yes, can the server reject this Command?

Putting it more strongly: `LiteralFound { tag: None }` during *command* parsing should be `unreachable!`.

`Option<T>` is there for ~my own sanity~ technical reason. If we would like to make this stronger, we would need to split `DecodeError` into `CommandDecodeError` and `ResponseDecodeError`. Let me try something out. Actually, this would resolve some other `unreachable!` cases.

jakoschiko:
Can't you simply preinitialize the `Tag` with a const default value?

duesee:
I think the better solution would be to split the `DecodeError`. The `tag` is not the only thing that doesn't make sense during response decoding. There is also #332 and currently `DecodeError` could also tell that we received a non-sync literal in a response which is not allowed.

duesee:
```rust
#[derive(Clone, Debug, Eq, PartialEq)]
pub enum CommandDecodeError<'a> {
    Incomplete,

    LiteralFound {
        tag: Tag<'a>,
        length: u32,
        mode: LiteralMode,
    },

    Failed,
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub enum ResponseDecodeError {
    Incomplete,

    LiteralFound {
        // There must be no `tag` here.
        // We keep the information about a literal because it allows the client to "discard" literal data silently.
        // I'll write something about this in the documentation.
        length: u32,
        // There must be no `mode` here.
    },

    Failed,
}
```

imap-codec/src/codec/decode.rs
```diff
@@ -92,120 +94,315 @@ impl<I> FromExternalError<I, base64::DecodeError> for IMAPParseError<I> {
...
+    /// More data is needed.
+    Incomplete,
+
+    /// Decoding failed.
+    Failed,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub enum DecodeError {
+pub enum CommandDecodeError<'a> {
```
duesee, line 112: `impl IntoBoundedStatic`.

imap-codec/src/codec/decode.rs
```diff
@@ -92,120 +94,315 @@ impl<I> FromExternalError<I, base64::DecodeError> for IMAPParseError<I> {
...
+    ///         Capability::LiteralPlus => /* Accept data right away. */
+    ///         Capability::LiteralMinus => {
+    ///             if literal_length <= 4096 {
+    ///                 /* Accept data right away. */
+    ///             } else {
+    ///                 /* Same as sync. */
+    ///             }
+    ///         }
+    ///     }
+    /// }
```
duesee, line 150:
```suggestion
    /// }
    ```
```

PR #337, opened 2023-08-05
reactor: Replace trait `Decode` with `Decoder`

jakoschiko:
Replaces the current trait `Decode` with the trait `Decoder` in order to improve lifetime issues.

I don't expect this PR to be merged, I just want to demonstrate my ideas. So, what's the problem with the current trait `Decode`?
```
pub trait Decode<'a>: Sized {
    fn decode(input: &'a [u8]) -> Result<(&'a [u8], Self), DecodeError>;
}
```
If I use this trait with type bounds, then I need to choose the lifetime in the wrong place:
```
fn foo<'a, T: Decode<'a>>() -> T { // The caller chooses a lifetime 'a
    let vec: Vec<u8> = b"foo".to_vec();
    let slice: &[u8] = &vec; // The slice's lifetime does not match lifetime 'a
    T::decode(slice); // Does not compile
}
```
This prevents me from writing useful abstractions based on `Decode`. So, here my proposed trait `Decoder`.
```
pub trait Decoder {
    type Item<'a>: Sized;

    fn decode<'a>(input: &'a [u8]) -> Result<(&'a [u8], Self::Item<'a>), DecodeError>;
}
```

It allows me to choose the lifetime at a later point in time:
```
fn foo<T: Decoder>() {
    let vec: Vec<u8> = b"foo".to_vec();
    let slice: &[u8] = &vec;
    T::decode(slice); // The lifetime is chosen here, now it compiles
}
```
List of changes:

- I replaced the trait `Decode` with `Decoder`
- I introduced the structs `CommandDecoder`, `ResponseDecoder`, etc. that implement `Decoder`
- I replaced the trait `DecodeStatic` with `StaticDecoder` that I was able to implement with less code duplication
- I fixed all usages of `Decode`

If you like the changes, we could do even more:

- For symmetry reasons we replace the trait `Encode` with `Encoder`
- We replace `CommandDecoder`, `ResponseDecoder`, etc. with `CommandCodec`, `ResponseCodec`, etc. that implement both `Decoder` and `Encoder`


duesee:
This makes so much more sense, thanks! :+1: (As discussed offline, let's talk about how, and when, to merge this later.)

duesee:
As discussed yesterday, I moved the decoder structs into codec and renamed them to `*Codec` already. I also removed the `StaticDecoder` and made it a default trait impl as you proposed. Is there anything left we want to do in this PR?

Two possible paths to continue: 1) Continue with `Encoder` or 2) #333.

jakoschiko:
I suggest this order: merge this, merge #333, add `Encoder`

PR #345, opened 2023-08-10
refactor: Rework `Encode`.

duesee:
* [x] Chore.
* [x] Simplify module hierachy.
* [x] Rework `Encode`.

@jakoschiko: Only the last commit is interesting.


imap-codec/src/codec/encode.rs
```diff
@@ -32,14 +36,25 @@ use imap_types::{
...
-    /// Create an [`Encoded`] for this message.
-    fn encode(&self) -> Encoded;
+#[cfg(feature = "ext_idle")]
+use crate::IdleDoneCodec;
+use crate::{AuthenticateDataCodec, CommandCodec, GreetingCodec, ResponseCodec};
+
+/// Encoder.
+///
+/// Implemented for types that know how to encode a specific IMAP message. See [implementors](trait.Encoder.html#implementors).
+pub trait Encoder {
```
duesee, line 46: Here is the changed trait.

PR #417, opened 2024-01-10
[DO NOT MERGE] Follow modifications done to imap-codec for Aerogramme 

superboum:
I am currently working on implementing CONDSTORE in Aerogramme. You can track my work [on the dedicated merge request thread](https://git.deuxfleurs.fr/Deuxfleurs/aerogramme/pulls/71). I am opening a PR so you can see which features I needed to add so the feature works. The code is ugly and is not meant to be merged.

Relevant RFC: [4466](https://datatracker.ietf.org/doc/html/rfc4466) + [RFC7162](https://datatracker.ietf.org/doc/html/rfc7162)

### DIFF
--- a/imap-codec/src/codec/encode.rs
+++ b/imap-codec/src/codec/encode.rs
@@ -439,7 +439,14 @@ impl<'a> EncodeIntoContext for CommandBody<'a> {
             }
             CommandBody::Check => ctx.write_all(b"CHECK"),
             CommandBody::Close => ctx.write_all(b"CLOSE"),
-            CommandBody::Expunge => ctx.write_all(b"EXPUNGE"),
+            CommandBody::Expunge { uid_sequence_set } => {
+                if let Some(seqset) = uid_sequence_set {
+                    ctx.write_all(b"UID EXPUNGE ")?;
+                    seqset.encode_ctx(ctx)
+                } else {
+                    ctx.write_all(b"EXPUNGE")
+                }
+            }
             CommandBody::Search {
                 charset,
                 criteria,
--- a/imap-codec/src/codec/encode.rs
+++ b/imap-codec/src/codec/encode.rs
@@ -55,7 +55,7 @@ use imap_types::{
         BasicFields, Body, BodyExtension, BodyStructure, Disposition, Language, Location,
         MultiPartExtensionData, SinglePartExtensionData, SpecificFields,
     },
-    command::{Command, CommandBody, StoreModifier},
+    command::{Command, CommandBody, StoreModifier, FetchModifier, SelectExamineModifier},
     core::{
         AString, Atom, AtomExt, Charset, IString, Literal, LiteralMode, NString, Quoted,
         QuotedChar, Tag, Text,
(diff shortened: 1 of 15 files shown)


superboum:
In the end, I would be interested to know what is the proper way to implement features in imap_codec for sure! But I would be more comfortable with a progressive approach. The first step seems to be that I have commits from you that are not in main. Should I discard them?

duesee:
> I have commits from you that are not in main. Should I discard them?

refactor!: Replace StatusV2 with Status -> already on main, discard
refactor(encoder)!: Add Fragment::AuthData -> discard (deleted from main)
feat(imap-{types,codec})!: Implement AuthenticateData::Cancel -> already on main, discard
quirk: mail.ru WIP -> already on main, discard

superboum:
Ok so now I'm in sync with main and I have a single commit that contain only my changes: allowing SELECT/EXAMINE/STORE to have modifiers.  Another FETCH modifier will come soon also ^^. Let me know if it's what you expect :-)

If everything is OK, the next step I envision to discuss with you is about your high-level guidelines to implement modifiers. Indeed, it seems that CONDSTORE introduce this novel concept of MODIFIERS in IMAP, concept that is described in a more generic way in [RFC4466](https://datatracker.ietf.org/doc/html/rfc4466). So should we aim at being generic here? Or stay specific? The generic approach seems to be limited by the fact that the exact syntax of the modifiers value is not given, or at least I have not understood it.

duesee:
> In the end, I would be interested to know what is the proper way to implement features in imap_codec for sure!

I don't know! 🙈 IMAP extensions are really more of "amendements"¹. It's difficult to reflect this in a library w/o breaking changes (or a very weird API).

This is our current approach:

When we add new extensions, we put them behind a feature flag starting with `ext_`. This feature flag has basically only the rule that everything should work when it's not used. You can use it ti add variants to an enum -- even when the enum is *not* `#[non_exhaustive]`. This would normally be a breaking change but given that we document this features as experimental, we try to give us freedom here to experiment with features before we stabilize them. The expectation is that we will could have a few, say, 5 or so, `ext_` features. And, the next time we do a major version, we'll "inline" the features into `imap-{types,codec}` (removing the feature flag.)

¹ I stole this sentence from Simon (go-imap).

duesee:
> What shape has the imap-type you envision for these modifiers?

I'm not confident in this part of the spec yet. Let me just "think aloud".

If I understand correctly, we have ...

```abnf
;; recommended overarching syntax for extensions
tagged-ext          = tagged-ext-label SP tagged-ext-val

;; Is a valid RFC 3501 "atom".
tagged-ext-label    = tagged-label-fchar *tagged-label-char

tagged-label-fchar  = ALPHA / "-" / "_" / "."

tagged-label-char   = tagged-label-fchar / DIGIT / ":"




tagged-ext-val      = tagged-ext-simple / "(" [tagged-ext-comp] ")"

tagged-ext-simple   = sequence-set / number

;; Extensions that follow this general
;; syntax should use nstring instead of
;; astring when appropriate in the context
;; of the extension.
;; Note that a message set or a "number"
;; can always be represented as an "atom".
;; An URL should be represented as
;; a "quoted" string.
tagged-ext-comp     = astring /
                      tagged-ext-comp *(SP tagged-ext-comp) /
                      "(" tagged-ext-comp ")"
```

... for the "generic approach", right?

superboum:
The weird part for me is:

> ;; Extensions that follow this general
;; syntax should use nstring instead of
;; astring when appropriate in the context
;; of the extension.

Which means that all extensions will not follow this syntax?! So I am not sure to understand the ins and the outs of implementing a generic approach that does not apply to all cases.

duesee:
> Which means that some extensions do not follow the general syntax?! So I am not sure to understand the ins and the outs of implementing a generic approach that does not apply to all cases.

Me neither. Accepting `NString` in case of `AString` would "hide" a possible `Atom("nil")`?

superboum:
Maybe we should not go for the generic modifier implementation? And only try to support modifiers for a selection of IMAP extensions like condstore? It seems that this generic specification is not very useful: another thing being that, while number are generally u32 in IMAP, CONDSTORE MODSEQ are u64, so when you use modifiers like CHANGEDSINCE XXX or UNCHANGED SINCE XXX, XXX is a u64.

imap-codec/src/codec/encode.rs
```diff
@@ -325,16 +325,34 @@ impl<'a> EncodeIntoContext for CommandBody<'a> {
...
-            CommandBody::Select { mailbox } => {
+            CommandBody::Select { mailbox, parameters } => {
                 ctx.write_all(b"SELECT")?;
                 ctx.write_all(b" ")?;
-                mailbox.encode_ctx(ctx)
+                mailbox.encode_ctx(ctx)?;
+                if let Some(p) = parameters {
+                    ctx.write_all(b" (")?;
+                    for atom in p.as_ref().iter() {
+                        atom.encode_ctx(ctx)?;
```
duesee, line 335: Just because I saw it: I think there is whitespace missing between the atoms? Maybe use `join_serializable`?

imap-codec/src/codec/encode.rs
```diff
@@ -325,16 +325,34 @@ impl<'a> EncodeIntoContext for CommandBody<'a> {
...
-            CommandBody::Examine { mailbox } => {
+            CommandBody::Examine { mailbox, parameters } => {
                 ctx.write_all(b"EXAMINE")?;
                 ctx.write_all(b" ")?;
-                mailbox.encode_ctx(ctx)
+                mailbox.encode_ctx(ctx)?;
+                if let Some(p) = parameters {
+                    ctx.write_all(b" (")?;
+                    for atom in p.as_ref().iter() {
+                        atom.encode_ctx(ctx)?;
```
duesee, line 350: Just because I saw it: I think there is whitespace missing between the atoms? Maybe use `join_serializable`?

superboum:
Thanks for your review, indeed I am not testing the client encoding part as I am not using it yet. Thanks also for the input on Option<NonEmptyVec> VS simply Vec<>, it makes sense :-) I think I will finish a first implementation on Aerogramme, test it, and then rewrite my code, and especially drop the generic approach of RFC4466 in favor of writing something specific to RFC7162. I will keep you updated here :-)

duesee:
Closing in favor of https://github.com/duesee/imap-codec/pull/439 :-)

PR #428, opened 2024-01-17
 refactor!,feat: Replace `NonEmptyVec` with `Vec1`

### LINKED ISSUE #427: refactor/question: Rename `NonEmptyVec` to `Vec1`?
There is a case now that requires two elements. If we do the rename, we can use `Vec1`, `Vec2`, ...

duesee:
Closes #427 

### TODO

* [x] Add a test for `Vec<T, N` for `N != 1`

### DIFF
--- a/imap-types/src/core.rs
+++ b/imap-types/src/core.rs
@@ -1399,31 +1399,40 @@ impl<'a> AsRef<str> for Charset<'a> {
     }
 }
 
-/// A `Vec` that always contains >= 1 elements.
+/// A [`Vec`] containing >= N elements.
 ///
-/// Some messages in IMAP require a list of at least one element. We encoded these situations in a
-/// non-empty vector type to not produce invalid messages.
+/// Some messages in IMAP require a list of *at least N* elements.
+/// We encode these situations with a specific vector type to not produce invalid messages.
 ///
-/// The `Debug` implementation equals `Vec` with an attached `+` at the end.
+/// Notes:
+///
+/// * `Vec<T, 0>` must not be used. Please use the standard [`Vec`] instead.
+/// * `Vec<T, 1>` must not be used. Please use the alias [`Vec1<T>`] instead.
 #[cfg_attr(feature = "bounded-static", derive(ToStatic))]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Clone, PartialEq, Eq, Hash)]
-pub struct NonEmptyVec<T>(pub(crate) Vec<T>);
+pub struct VecN<T, const N: usize>(pub(crate) Vec<T>);
 
-impl<T> Debug for NonEmptyVec<T>
+impl<T, const N: usize> Debug for VecN<T, N>
 where
     T: Debug,
 {
     fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
         self.0.fmt(f)?;
-        write!(f, "+")
+        match N {
+            0 => write!(f, "*"),
+            1 => write!(f, "+"),
+            _ => write!(f, "{{{},}}", N),
+        }
     }
 }
 
-impl<T> NonEmptyVec<T> {
+impl<T, const N: usize> VecN<T, N> {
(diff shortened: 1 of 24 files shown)


imap-types/src/core.rs
```diff
@@ -1475,12 +1469,34 @@ impl<T> IntoIterator for NonEmptyVec<T> {
...
+    }
+}
+
+/// A `Vec` that always contains >= 2 elements.
+///
+/// Some messages in IMAP require a list of at least one element. We encoded these situations in a
+/// non-empty vector type to not produce invalid messages.
+///
+/// The `Debug` implementation equals `Vec` with an attached `{2,}` at the end.
+pub type Vec2<T> = VecN<T, 2>;
```
jakoschiko, line 1498: Hm, I'm not sure whether I like the name. Based on the name I'd assume that it has exactly 2 elements. Like the `Vec2` from the `glam` crate.

Alternatives:
- `Min2Vec`
- `Vec2Plus`
- `VecAtLeast2`
- `Vec2OrMoah`

imap-types/src/core.rs
```diff
@@ -1399,31 +1399,31 @@ impl<'a> AsRef<str> for Charset<'a> {
...
-///
-/// Some messages in IMAP require a list of at least one element. We encoded these situations in a
-/// non-empty vector type to not produce invalid messages.
-///
-/// The `Debug` implementation equals `Vec` with an attached `+` at the end.
 #[cfg_attr(feature = "bounded-static", derive(ToStatic))]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 #[derive(Clone, PartialEq, Eq, Hash)]
-pub struct NonEmptyVec<T>(pub(crate) Vec<T>);
+pub struct VecN<T, const N: usize>(pub(crate) Vec<T>);
```
jakoschiko, line 1405: It looks like you couldn't decide whether you make this type public or not.

I'd would prefer to document `VecN` extensively and replace the doc of `Vec1` and `Vec2` with a short comment.

imap-types/src/core.rs
jakoschiko, line 1: `impl<T; const N: usize> From<[T; N]> for VecN<T, N>` would also be useful.

imap-types/src/core.rs
```diff
@@ -1475,12 +1469,34 @@ impl<T> IntoIterator for NonEmptyVec<T> {
...
+
+impl<T> From<T> for Vec1<T> {
+    fn from(value: T) -> Self {
+        VecN(vec![value])
+    }
+}
+
+/// A `Vec` that always contains >= 2 elements.
+///
+/// Some messages in IMAP require a list of at least one element. We encoded these situations in a
```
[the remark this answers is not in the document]
duesee:
```suggestion
/// Some messages in IMAP require a list of at least two elements. We encoded these situations in a
```

imap-types/src/core.rs
duesee: I see what you mean but still lean towards keeping `Vec2`. nom [uses](https://docs.rs/nom/latest/nom/multi/fn.separated_list1.html), e.g., `separated_list{0,1}` with the meaning >= 0 and >= 1. You may need to look it up, but given that I expect people to be often exposed to `Vec1`, maybe it's okay they need to look it up once? imap-types already has quite some "typing required" (and rightward drift). So -- while usually a bad argument! -- it makes a short name like `Vec2` hard to beat here... What do you think?

imap-types/src/core.rs
duesee: Hehe, that's exactly what happend :-) Couldn't publicly alias a private type (of course). Good catch!

imap-types/src/core.rs
jakoschiko: Same for Vec1 and VecN. But no strong feelings from my side.

imap-types/src/core.rs
jakoschiko: doc(hidden) would be an option. But I like VecN and would prefer to keep it and document it properly.

PR #440, opened 2024-02-09
feat: Implement `Clone` and `PartialEq` for codec types

### LINKED ISSUE #438: Implement `Clone` for codec types
I would like to have a `Clone` implementation for types like `CommandCodec`, etc. Any objects?

jakoschiko:
If I understand correctly the codec structs (e.g. `CommandCodec`) might contain options in the future that effect how messages are encoded/decoded. It's useful to implement `Clone` and `PartialEq` for such option structs. This PR does this.

Resolves #438


duesee:
Thanks!

PR #468, opened 2024-03-16
quirk: deal with Exchange's excessive space

nbdd0121:
Microsoft Exchange responds STATUS with a trailing space, which is not RFC compliant.

This is similarly worked around in Ruby's imap library: https://bugs.ruby-lang.org/issues/13649


duesee:
Perfect. Thank you very much, @nbdd0121!

PR #477, opened 2024-04-06
perf: add more benchmarks

duesee:


imap-codec/benches/greeting.rs
```diff
@@ -0,0 +1,70 @@
...
+                b.iter(|| {
+                    serialize(&codec, &object);
+                })
+            },
+        );
+
+        let input = serialize(&codec, &object);
+        c.bench_function(format!("bench_greeting_parse_{instance}").as_str(), |b| {
+            b.iter(|| {
+                parse(&codec, black_box(&input[..]));
```
jakoschiko, line 27: Because of the `;` the return value will not be returned to criterion. I think this might prevent the expression from being evaluated.

PR #479, opened 2024-04-06
build: add initial justfile

duesee:
The `justfile` is already close to what is in the GitHub Actions CI. A few things are missing and some are untested, though. My plan is to gradually replace the GitHub Actions CI with mere `just <job>` calls. This should allow to quickly run jobs locally and become more independent of GitHub.


justfile
```diff
@@ -0,0 +1,155 @@
+export RUSTFLAGS := "-D warnings"
+export RUSTDOCFLAGS := "-D warnings"
```
jakoschiko, line 2: Are these variables used for all just recipes? Is this necessary?

justfile
```diff
@@ -0,0 +1,155 @@
...
+
+
+
+[private]
+default:
+	just -l --unsorted
```
jakoschiko, line 8: Would it make sense to bind the default recipe to a more frequently used command? Something like "I always want to run this before I commit".

justfile
```diff
@@ -0,0 +1,155 @@
...
+
+
+
+# Install rust 1.65, cargo-deny, cargo-fuzz, cargo-hack, and cargo-semver-checks
+install: install_rust_1_65 \
```
jakoschiko, line 13: That's good

justfile
jakoschiko, line 1: Are the empty lines in this file intentional?

justfile
```diff
@@ -0,0 +1,155 @@
...
+
+
+
+# Check syntax, formatting, clippy, deny, ...
+quick: (_quick ""               ""         ) \
+       (_quick ""               "--release") \
+       (_quick "--all-features" ""         ) \
+       (_quick "--all-features" "--release")
+
+_quick features mode: (cargo_check features mode) cargo_fmt (cargo_clippy features mode) cargo_deny
```
jakoschiko, line 47: Maybe you can rewrite it to `[private] quick_impl` instead of using the old private method naming convention with `_`. It might be less confusing for beginners (like me).

justfile
```diff
@@ -0,0 +1,155 @@
...
+
+
+
+# Check extensively (required for PR)
+pr: (_pr ""               ""         ) \
+    (_pr ""               "--release") \
+    (_pr "--all-features" ""         ) \
+    (_pr "--all-features" "--release")
+
+_pr features mode: quick semver (cargo_hack mode) (cargo_test features mode)
```
jakoschiko, line 80: Same here

justfile
```diff
@@ -0,0 +1,155 @@
...
+
+# Install rust 1.65, cargo-deny, cargo-fuzz, cargo-hack, and cargo-semver-checks
+install: install_rust_1_65 \
+         install_cargo_deny \
+	 install_cargo_fuzz \
+	 install_cargo_hack \
+	 install_cargo_semver_checks 
+
+[private]
+install_rust_1_65:
```
jakoschiko, line 20: Why 1.65? Is this the MSRV? If yes, please add a comment.

And would it make sense to use `+1.65` for all cargo commands? E.g. `cargo +1.65`.

justfile
jakoschiko, line 13: What about nightly rust for `cargo +nightly fmt` and `cargo +nightly fuzz`?

justfile
duesee: Not sure... let's remove it when it starts to get in the way.

PR #494, opened 2024-05-21
Structure Aware Tag Fuzz Target

spellfish:
Adds Basic Structure Aware Fuzzing for tag_imap:

- TagInput: A struct representing the input for the fuzz target.
- Arbitrary Implementation: This allows TagInput to be generated with random values.
- Conversion to Bytes: Converts the input string to bytes.
- Parsing: Attempts to parse the input bytes as an IMAP tag using tag_imap.
- Currently does not specify additional checks in the even of a parsing failure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -505,6 +505,7 @@ dependencies = [
 name = "imap-codec-fuzz"
 version = "0.0.0"
 dependencies = [
+ "arbitrary",
  "imap-codec",
  "imap-types",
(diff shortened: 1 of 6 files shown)


duesee (approved):
Thank you very much @spellfish! Also for all your other great work! :-)

Sorry that the review took longer... I'm about to release imap-codec 2.0.0-alpha-1 and wanted to feature-gate the fuzz API first.

PR #501, opened 2024-06-08
fix: Handle different line endings and EOF in examples

HenningHolmDE:
The `parse_*` examples for imap-codec were not running on Windows due to an incorrect conversion of an already existing `CRLF` line ending to an erroneous `CRCRLF` line ending.
This fixes the issue by ensuring a `CRLF` line ending in all possible cases.

Also, EOF of stdin is now handled and no longer results in an endless loop (e.g. `echo -n "* OK Hello" | cargo run -p imap-codec --example parse_greeting`).

### DIFF
--- a/imap-codec/examples/common/common.rs
+++ b/imap-codec/examples/common/common.rs
@@ -24,7 +24,8 @@ pub fn read_more(buffer: &mut Vec<u8>, role: Role) {
 
     let line = read_line(prompt, role);
 
-    if line.trim() == "exit" {
+    // If `read_line` produces an empty string, standard input has been closed.
+    if line.is_empty() || line.trim() == "exit" {
         println!("Exiting.");
         std::process::exit(0);
     }
@@ -44,6 +45,11 @@ fn read_line(prompt: &str, role: Role) -> String {
     std::io::stdin().read_line(&mut line).unwrap();
 
     print!("{RESET}");
+
+    // If `Stdin::read_line` produces an empty string, standard input has been closed.
+    if line.is_empty() {
+        return line;
+    }
 
     // Ensure `CRLF` line ending of resulting string.
     // Line ending of `line` can be one of:
--- a/imap-codec/examples/common/common.rs
+++ b/imap-codec/examples/common/common.rs
@@ -45,5 +45,16 @@ fn read_line(prompt: &str, role: Role) -> String {
 
     print!("{RESET}");
 
-    line.replace('\n', "\r\n")
+    // Ensure `CRLF` line ending of resulting string.
+    // Line ending of `line` can be one of:
+    // - `CRLF` on Windows
+    // - `LF` on Unix-like
+    // - none when EOF of standard input is reached
+    if line.ends_with("\r\n") {
+        return line;
+    }
+    if line.ends_with('\n') {
+        line.pop();
+    }
+    line + "\r\n"
 }


duesee:
Uhuuuu, good catch! I think we can do better: I think it's possible to require features for examples, right?

We could enable https://github.com/duesee/imap-codec/blob/65e99b5b68bf1f015ec524ffae1452b23f2b9d29/imap-codec/Cargo.toml#L43 and make it work w/o changing the data. Would simplify the example.

What do you think?

imap-codec/examples/common/common.rs
```diff
@@ -45,5 +45,12 @@ fn read_line(prompt: &str, role: Role) -> String {
 
     print!("{RESET}");
 
-    line.replace('\n', "\r\n")
+    // Ensure CRLF line endings
+    // (could be LF on Unix or CRLF on Windows)
+    line.lines()
```
duesee, line 50: I... think I see what happens here. But maybe something like... 

```rust
if !line.ends_with("/r/n) {
    line.replace()
}
```

... would be easier to understand? (In case the feature quirk idea turns out to introduce too much hassle.)

HenningHolmDE: Okay, let's dig deeper into this to find the optimal solution ...

The current implementation `line.replace('\n', "\r\n")` will change _all `LF` line endings_ in `line` to `CRLF` (also, unfortunately, incorrectly change `CRLF` to `CRCRLF`). My proposed change would change _all line endings_ (`LF` or `CRLF`) in `line` to `CRLF`.

However, looking into the documentation of [BufRead::read_line](https://doc.rust-lang.org/stable/std/io/trait.BufRead.html#method.read_line) there should only be one `LF` at the end of the string (which might belong to a `CRLF`) - or none on EOF if stdin has been closed. Therefore, we could indeed simplify this to only check the end of the string.
(As is turns out, EOF is currently also not correctly handled and should probably be interpreted as `exit` instead of infinitely looping on empty input. 😉 I will add a second commit fixing this as well, after we have decided on an implementation here.)

Basically, in this location we want to do something like this to handle the three possible cases:
```rust
    if line.ends_with("\r\n") {
        return line;
    }
    if line.ends_with("\n") {
        line.pop();
    }
    line + "\r\n"
```

Do we care about trailing spaces to be preserved? Otherwise, we could also do something like:
```rust
line.trim_end().to_string() + "\r\n"
```
However, in contrast to the previous snippets, this would copy the string once after trimming.

What are your thoughts on how to proceed here?

HenningHolmDE:
> Uhuuuu, good catch! I think we can do better: It's possible to require features for examples.
> 
> We could enable ...
> 
> https://github.com/duesee/imap-codec/blob/65e99b5b68bf1f015ec524ffae1452b23f2b9d29/imap-codec/Cargo.toml#L43
> 
> ... to make the examples work w/o changing the data.
> 
> What do you think?

I don't think it the best way to fix an terminal input issue in the example by forcing the user to enable the quirk feature and relaxing the parser to deal with the input. We know what range of input to expect from `Stdin::read_line` in different situations and should be able to handle it locally in `common::read_line`.

imap-codec/examples/common/common.rs
duesee: Great analysis, thanks! I like the first solution you propose!

> Do we care about trailing spaces to be preserved?

Yes. You could for example have a greeting ...

```
* OK Hello___\r\n
```

... that should become ...

```rust
Greeting {
  kind: Ok,
  code: None,
  text: "Hello___",
}
```

... with "_" meaning space here.

duesee:
Agreed 👍🏻

imap-codec/examples/common/common.rs
HenningHolmDE: Thanks for your input. I have now implemented the first solution and also added a fix for handling EOF correctly.

PR #503, opened 2024-06-14
fix: Resolve conflicting macro definition

HenningHolmDE:
Previously, there were two macros named `impl_try_from` in scope in `mailbox.rs` (from import and local definition). While the build correctly resolved both macros, rust-analyzer did not and complained about using the imported macro.
Renaming the local macro resolves this ambiguity and makes rust-analyzer happy.

### DIFF
--- a/imap-types/src/mailbox.rs
+++ b/imap-types/src/mailbox.rs
@@ -235,7 +235,7 @@ impl<'a> MailboxOther<'a> {
     }
 }
 
-macro_rules! impl_try_from {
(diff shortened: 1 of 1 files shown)


duesee (approved):
Huch... I would have guesses this doesn't compile... Thanks for the fix!

HenningHolmDE:
> Huch... I would have guessed this doesn't compile...

Yeah, would have thought as well. Learning something new every day. 🧑‍🎓 

> Thanks for the fix!

You're welcome!

PR #504, opened 2024-06-14
fix(misuse): Add validation to deserialization

### LINKED ISSUE #502: misuse: Deserialization into incorrect types
Currently, types created through `deserialize` are not validated. Thus, it is possible to create incorrect values, e.g.:
```rust
#!/usr/bin/env rust-script
//! Try to deserialize invalid input into a `core::Text` type
//!
//! ```cargo
//! [dependencies]
//! imap-types = { git = "https://github.com/duesee/imap-codec.git", features = ["serde"] }
//! serde_json = "1.0"
//! ```

use imap_types::core::Text;

fn main() {
    let valid_input = r#""Hello, world!""#;
    let invalid_input = r#""Hello,\rworld!""#;

    let text = serde_json::from_str::<Text>(valid_input)
        .expect("valid input should deserialize successfully");
    assert_eq!(text.as_ref(), "Hello, world!");

    let err = serde_json::from_str::<Text>(invalid_input)
        .expect_err("invalid input should not deserialize successfully");
    assert_eq!(
        err.to_string(),
        r"Validation failed: Invalid byte b'\x0d' at index 6"
    );

    println!("Success.")
}
```

```bash
$ rust-script deserialize_invalid_text.ers    
thread 'main' panicked at ...\deserialize_invalid_text.ers:21:10:
invalid input should not deserialize successfully: Text("Hello,\rworld!")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

At least for `core::Text`, it is possible to fix the issue using [`#[serde(try_from = "String")]`](https://serde.rs/container-attrs.html#try_from).

I will try to provide a corresponding PR in the next days.

HenningHolmDE:
Previously, correctness through the existing validation functions was
not enforced during deserialization.
This is now achieved by going through `TryFrom` implementations or
specific field deserialization functions. Thus, it should no longer be
possible to create invalid instances by deserialization.

This fixes #502.

### DIFF
--- a/imap-types/src/datetime.rs
+++ b/imap-types/src/datetime.rs
@@ -11,6 +11,7 @@ use serde::{Deserialize, Serialize};
 use crate::datetime::error::{DateTimeError, NaiveDateError};
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(try_from = "chrono::DateTime<FixedOffset>"))]
 #[derive(Clone, Eq, PartialEq, Hash)]
 pub struct DateTime(chrono::DateTime<FixedOffset>);
 
@@ -96,6 +97,7 @@ impl ToBoundedStatic for DateTime {
 }
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(try_from = "chrono::NaiveDate"))]
 #[derive(Clone, Eq, PartialEq, Hash)]
 pub struct NaiveDate(chrono::NaiveDate);
 
@@ -267,5 +269,46 @@ mod tests {
             println!("{:?}", got.clone().unwrap_err());
             assert_eq!(expected, got.unwrap_err());
         }
+    }
+
+    #[cfg(feature = "serde")]
+    #[test]
+    fn test_deserialization_date_time() {
+        let valid_input = r#""2015-05-15T11:22:33+01:00""#;
+        let invalid_input = r#""+12015-05-15T11:22:33+01:00""#;
+
+        let date_time = serde_json::from_str::<DateTime>(valid_input)
+            .expect("valid input should deserialize successfully");
+        assert_eq!(
+            date_time,
+            DateTime(
+                chrono::FixedOffset::east_opt(3600)
+                    .unwrap()
+                    .with_ymd_and_hms(2015, 5, 15, 11, 22, 33)
+                    .unwrap(),
+            ),
+        );
+
+        let err = serde_json::from_str::<DateTime>(invalid_input)
+            .expect_err("invalid input should not deserialize successfully");
+        assert_eq!(err.to_string(), r"expected `0 <= year <= 9999`, got 12015");
+    }
+
+    #[cfg(feature = "serde")]
+    #[test]
+    fn test_deserialization_naive_date() {
+        let valid_input = r#""2015-05-15""#;
+        let invalid_input = r#""+12015-05-15""#;
+
+        let naive_date = serde_json::from_str::<NaiveDate>(valid_input)
+            .expect("valid input should deserialize successfully");
+        assert_eq!(
+            naive_date,
+            NaiveDate(chrono::NaiveDate::from_ymd_opt(2015, 5, 15).unwrap()),
+        );
+
+        let err = serde_json::from_str::<NaiveDate>(invalid_input)
+            .expect_err("invalid input should not deserialize successfully");
+        assert_eq!(err.to_string(), r"expected `0 <= year <= 9999`, got 12015");
     }
 }
--- a/imap-types/src/core.rs
+++ b/imap-types/src/core.rs
@@ -1327,13 +1327,13 @@ impl TryFrom<char> for QuotedChar {
 /// looking at the `search` rule. (See [#266](https://github.com/duesee/imap-codec/issues/266).)
 ///
 /// ```abnf
-/// search          = "SEARCH" [SP "CHARSET" SP astring] 1*(SP search-key)
-/// ;                                           ^^^^^^^
-/// ;                                           |
-/// ;                                           `Charset`
-//                     ; CHARSET argument to MUST be registered with IANA
+/// search = "SEARCH" [SP "CHARSET" SP astring] 1*(SP search-key)
+///            ;                       ^^^^^^^
+///            ;                       |
+///            ;                       `Charset`
+///            ; CHARSET argument to MUST be registered with IANA
 /// ```
-/// 
+///
 /// So, it seems that it should be an `AString`. However the IMAP standard also points to ...
 /// ```abnf
 /// mime-charset       = 1*mime-charset-chars
--- a/imap-types/src/core.rs
+++ b/imap-types/src/core.rs
@@ -98,6 +98,7 @@ use crate::extensions::binary::Literal8;
 /// ```
 #[cfg_attr(feature = "bounded-static", derive(ToStatic))]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(try_from = "String"))]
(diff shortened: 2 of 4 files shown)


HenningHolmDE:
@duesee I just read once again through your reply in the issue and noticed that I overlooked your hint to `CommandContinuationRequestBasic`, which has no `validate` method but a validation in its `new` method. I will take a look at that tomorrow.

(Also I will look into the strange non-diff the failing check is complaining about.)

jakoschiko:
You are using `#[serde(try_from = "String")]`. As far as I can tell this will create an intermediate copy and do some unnecessary unicode checks. Did you try `#[serde(try_from = "&'a [u8]")]`? I was not able to compile your branch with my suggestion, but [it works in playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=68260bf9be66d2f05628315999ec8614).

HenningHolmDE:
@jakoschiko Thanks for pointing this out. I will try to address the issues in your comment one by one:

> You are using #[serde(try_from = "String")]. [...]

Hopefully, I only added deserialization through  `String` in cases, where the underlying type is some type of string (e.g. `Cow<'a, str>`). If you found any occurence where I went through a `String` on the way to a byte type (e.g. `Cow<'a, [u8]>`), please let me know. That would definetely be incorrect.

> [...] As far as I can tell this will create an intermediate copy [...]

I would agree that deserializing into `String` will copy the data from the deserializer. After that, moving through `TryFrom` into `Cow` should not result into creating an additional copy. Let's check this [in the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3e09fb9ec36e92352b42eda3582be44f).

However, if you are suggesting that previously, data was not copied, this seems not to be the case as deserialization into a `Cow` always results in a `Cow::Owned`, thus copying the data. (see https://github.com/serde-rs/serde/issues/1852)
There seem to be ways to mitigate this (e.g. using [`#[serde(borrow)]`](https://serde.rs/field-attrs.html#borrow) or introducing [`serde_with::BorrowCow`](https://docs.rs/serde_with/latest/serde_with/struct.BorrowCow.html)). But when I tried this out briefly, I ended up sprinkling `borrow` all over the place and yet was not able to make things work without copying when used in combination with `try_from`.
There is probably some way of doing this, but would suggest adding an issue to tackle this optimization at a later point in time. What do you think?

> [...] and do some unnecessary unicode checks. [...]

Where would you expect additional unicode checks? On the way to the first Rust string type in the conversion, there will always be a unicode check to ensure Rust's UTF-8 guarantees. However, when converting between string types, there should be no additional checks (e.g. when moving the `String` into a `Cow<'a, str>`). 

> [...] Did you try #[serde(try_from = "&'a [u8]")]? I was not able to compile your branch with my suggestion, but [it works in playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=68260bf9be66d2f05628315999ec8614).

I tried using `&'a str` in combination with `#[serde(borrow)]`, but as stated above, unfortunately, I was not able to easily get rid of any copying this way in the actual code.

HenningHolmDE:
@duesee I'm a bit lost with the Coveralls failure.

In the report, there seem to be arbitrarily sprinkled coverage misses in [empty lines](https://coveralls.io/builds/68118692/source?filename=imap-types%2Fsrc%2Fmailbox.rs#L110) and [comments](https://coveralls.io/builds/68118692/source?filename=imap-types%2Fsrc%2Fmailbox.rs#L176). I first thought that might have something to do with mixing in old coverage information into the report, but even with the changes to the coverage report generation we merged today, this is still the case.

Also, when I look at the coverage reports generated locally, the PR seems to only improve things in the files critizied by the tool and the report generated by `grcov` makes actual sense.

For the base report of main, everything seems to be fine and comparable to the local report.

Any ideas on this?

jakoschiko:
> Hopefully, I only added deserialization through `String` in cases, where the underlying type is some type of string (e.g. `Cow<'a, str>`). If you found any occurence where I went through a `String` on the way to a byte type (e.g. `Cow<'a, [u8]>`), please let me know. That would definetely be incorrect.

My bad. I thought that types like `Atom` were wrapping `&[u8]`, not `&str`.

> However, if you are suggesting that previously, data was not copied, this seems not to be the case as deserialization into a `Cow` always results in a `Cow::Owned`, thus copying the data.

The previous serialization was also non-optimal (probably by accident). It should be possible to have an implementation that does not copy the bytes. E.g. the `TryFrom` implementation for `Atom` is using `Cow::Borrowed`:

```rust
impl<'a> TryFrom<&'a str> for Atom<'a> {
    type Error = ValidationError;

    fn try_from(value: &'a str) -> Result<Self, Self::Error> {
        Self::validate(value)?;

        Ok(Self(Cow::Borrowed(value)))
    }
}
```

> There is probably some way of doing this, but would suggest adding an issue to tackle this optimization at a later point in time. What do you think?

Sure, if it's non-trivial to fix it, let's do it later.

> Where would you expect additional unicode checks? On the way to the first Rust string type in the conversion, there will always be a unicode check to ensure Rust's UTF-8 guarantees. 

Again, my bad. I'm a little bit surprised that @duesee decided to use `&str` instead of `&[u8]`. I far as I know we don't need to uphold Rust's UTF-8 guarantees here.

> I tried using `&'a str` in combination with `#[serde(borrow)]`, but as stated above, unfortunately, I was not able to easily get rid of any copying this way in the actual code.

This confuses me. Because for the `TryFrom` implementation I mentioned above I think this should be possible without much effort.

HenningHolmDE:
> I'm a little bit surprised that @duesee decided to use &str instead of &[u8]. I far as I know we don't need to uphold Rust's UTF-8 guarantees here.

When writing the previous comment, I gave this some more thought and I'm actually also not sure if assuming this is correct. I have created #506 for discussing this further.

> > I tried using &'a str in combination with #[serde(borrow)], but as stated above, unfortunately, I was not able to easily get rid of any copying this way in the actual code.
> 
> This confuses me. Because of the TryFrom implementation I mentioned above I think this should be possible without much effort.

Unfortunately, your playground example simplifies things a bit by not using `Cow` and therefor not requiring `#[serde(borrow)]`.
However, I'm not sure what I observed a few days ago, but trying it out now [in the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fec5ab60771799e346b1e4aa1ce3aee6), the combination indeed seems to work.
Anyways, there is still the point of having to add `#[serde(borrow)]` all the way up the type tree for which I'm not sure of the consequences. So I would rather not throw this improvement into this PR as well and have created #507 to keep track of the topic.

duesee:
Regarding the coverage: Maybe, maaaybe, we screwed up the GitHub Actions cache when porting everything to `just` a few months ago. I had this feeling already that something is flaky... If your local coverage is sane, I would suggest not blocking on this. I'll open an issue to take a closer look.

imap-types/src/datetime.rs
```diff
@@ -11,6 +11,7 @@ use serde::{Deserialize, Serialize};
 use crate::datetime::error::{DateTimeError, NaiveDateError};
 
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde", serde(try_from = "chrono::DateTime<FixedOffset>"))]
```
duesee, line 14: I'm not sure I understand this. Is this to get rid of the newtype? Is serde handling this automatically with `try_from`?

duesee (approved):
Thanks! This looks good to me. I'll file two issues:

* Check coverage in CI, and
* Fuzz-test serde (de)serialization (when serde feature is used)

@jakoschiko Do you think this can already be merged?

HenningHolmDE:
> Regarding the coverage: Maybe, maaaybe, we screwed up the GitHub Actions cache when porting everything to `just` a few months ago. I had this feeling already that something is flaky... If your local coverage is sane, I would suggest not blocking on this. I'll open an issue to take a closer look.

#511 should fix that. I am happy to wait for that one to be merged first and see this PR not reducing the coverage after rebasing.

duesee:
Awesome work! Thanks! :-)

PR #505, opened 2024-06-16
build: Improve coverage reporting

HenningHolmDE:
- remove coverage artifacts before running the tests as well (otherwise, a failing test run may leave coverage artifacts in place)
- keep all coverage related files in `target/coverage`
- add creation of HTML report for local development use

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -6,12 +6,14 @@ on:
       - "**.rs"
       - "**.toml"
       - ".github/workflows/**"
+      - "justfile"
   pull_request:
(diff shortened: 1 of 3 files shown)


duesee:
Thanks, this looks very good to me. I believe the reason GitHub doesn't pick up the jobs is because we have a job filter in place. Could you add `justfile ` to the filter in https://github.com/duesee/imap-codec/blob/main/.github/workflows/main.yml#L8 (and below) to make the CI pick this changes up?

duesee (approved):
Thanks!

PR #511, opened 2024-06-18
fix(build): Use clean build for coverage test run

### LINKED ISSUE #508: Review CI cache (due to potentially flaky coverage)

HenningHolmDE:
Previously, coverage data could be messed up by build artifacts restored from the GitHub cache. While it is not yet cleared up how this actually happens, removing all build data before executing the coverage test run seems to resolve this issue.

This fixes #508.

### DIFF
--- a/justfile
+++ b/justfile
@@ -139,9 +139,9 @@ bench_against_main:
 
 # Measure test coverage
 coverage: install_rust_llvm_tools_preview install_cargo_grcov
-    mkdir -p target/coverage
-    # Remove old profiling information and coverage reports
-    rm -rf target/coverage/*
+    # Old build artifacts seem to be able to mess up coverage data (see #508),
+    # removing everything in `target` seems to be the easiest fix for this.
+    rm -rf target/*
     # Run instrumented tests to generate coverage information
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -139,9 +139,9 @@ bench_against_main:
 
 # Measure test coverage
 coverage: install_rust_llvm_tools_preview install_cargo_grcov
-    mkdir -p target/coverage
-    # Remove old profiling information and coverage reports
-    rm -rf target/coverage/*
+    # Old build artifacts seem to be able to mess up coverage data (see #508),
```
duesee, line 142: First thought "Could we use `cargo clean`. Second thought: "Ah, target is needed".

duesee:
Thanks, happy to merge (and curiously looking forward to the rebase :-))

justfile
HenningHolmDE: Actually, it is not needed, as it is recreated when running `cargo test` afterwards (also, any non-existing path in `LLVM_PROFILE_FILE` is automaically created).

My first thought was using `cargo clean` as well, then I thought "why calling a tool for cleaning a directory?". Would you rather have gone with using Cargo?

Issue #359, opened 2023-09-26
feat: Python bindings.


duesee:

duesee:
Hey Henning! I'd like to assign you this issue. The way this works in GiHub is that you need to write some comment here -- content is irrelevant. Then, GitHub allows me to assign this issue to you.

So... What's up? :-)

HenningHolmDE:
Sure, go ahead.

HenningHolmDE:
@duesee What do you think about splitting/renaming this issue to more actionable issues for better representing the progress towards the milestone?

duesee:
> @duesee What do you think about splitting/renaming this issue to more actionable issues for better representing the progress towards the milestone?

I would be very much in favor of it. This was kind of an "reminder" only but having more structure would be awesome.

HenningHolmDE:
Now that #517 has landed and Python bindings are somewhat on their way, we can probably close this generic issue and track everything else separately. What do you think?

PR #513, opened 2024-06-19
Create project for python bindings

HenningHolmDE:
- This creates a new `maturin` template project for `PyO3` in which the Python bindings for imap-codec will be implemented.
- It also adds basic Python unit tests for the example function `sum_to_string` of the template project that is available from the compiled Rust code.
- By using the new `bindings` just recipe running build and test of the Python bindings can be triggered.
- A new GitHub job for automatically running the build in CI is added to the workflow.

This is part of #359.

### DIFF
--- a/bindings/imap-codec-python/Cargo.toml
+++ b/bindings/imap-codec-python/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "imap-codec-python"
+description = "Python bindings for imap-codec"
+version = "0.1.0-dev1"
+license = "MIT OR Apache-2.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+name = "imap_codec"
+crate-type = ["cdylib"]
+# required as this lib name is identical to `imap-codec`
+doc = false
+
+[dependencies]
+pyo3 = "0.21.2"
--- a/deny.toml
+++ b/deny.toml
@@ -1,6 +1,6 @@
 [sources]
 unknown-registry = "deny"
-unknown-git      = "deny"
+unknown-git = "deny"
 allow-git = [
     # Used in benchmarking
     "https://github.com/stalwartlabs/mail-server?rev=53f0222f308b3e844c158fc0e603d10361da3c63",
@@ -9,7 +9,13 @@ allow-git = [
 ]
 
 [licenses]
-allow = [ "Apache-2.0", "MIT", "BSD-3-Clause", "Unicode-DFS-2016" ]
+allow = [
+    "Apache-2.0",
+    "Apache-2.0 WITH LLVM-exception",
+    "MIT",
+    "BSD-3-Clause",
+    "Unicode-DFS-2016",
+]
 
 [licenses.private]
 ignore = true
--- a/deny.toml
+++ b/deny.toml
@@ -4,6 +4,8 @@ unknown-git      = "deny"
 allow-git = [
     # Used in benchmarking
     "https://github.com/stalwartlabs/mail-server?rev=53f0222f308b3e844c158fc0e603d10361da3c63",
+    # Used for patching `pyo3-build-config` (see #516)
+    "https://github.com/PyO3/pyo3/?rev=2c205d4586bef8f839eb6a55eb91b905074ddce9",
 ]
 
 [licenses]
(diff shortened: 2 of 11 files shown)


HenningHolmDE:
@duesee This PR contains the infrastructure changes for the Python bindings to be implemented. I'm happy to discuss any suggestions on doing things differently.

justfile
```diff
@@ -137,6 +137,16 @@ bench_against_main:
     cargo bench -p imap-codec-bench
     rm -rf target/bench_tmp
 
+# Build and test bindings
+bindings: bindings_python
+
+# Build and test Python bindings
+bindings_python: install_python_maturin
+    rm -rf target/wheels/*
```
duesee, line 145: question: Can this fail when there is no `target` directory yet?

duesee (commented):
Thanks! Looks very good to me. I left a few questions. I cannot test right now but will the bindings job work after a `cargo clean`. Do we expect some job order?

Suggesstion for future: Use [`black`](https://github.com/psf/black) and maybe even [`mypy`](https://mypy-lang.org/) in CI? What do you think? :-)

deny.toml
```diff
@@ -1,13 +1,21 @@
 [sources]
 unknown-registry = "deny"
-unknown-git      = "deny"
```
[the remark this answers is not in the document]
HenningHolmDE: Oh, I seem to have missed removing one auto-formating unrelated to my changes.

deny.toml
```diff
@@ -1,13 +1,21 @@
...
+unknown-git = "deny"
 allow-git = [
     # Used in benchmarking
     "https://github.com/stalwartlabs/mail-server?rev=53f0222f308b3e844c158fc0e603d10361da3c63",
+    # Used for patching `pyo3-build-config`
```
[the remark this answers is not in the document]
HenningHolmDE: Good point. I will do that.

justfile
HenningHolmDE: I don't think so. From the `rm` man page: 
```
       -f, --force
              ignore nonexistent files and arguments, never prompt
```

bindings/imap-codec-python/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "imap-codec-python"
+description = "Python bindings for imap-codec"
+version = "2.0.0-alpha.1"
```
HenningHolmDE, line 4: @duesee I gave this some thought since yesterday. Maybe, it is not a good idea to use the same numbering scheme for the binding packages. Otherwise, every breaking change in the bindings would force us to bump the major version number for all crates.
At some point, when the Python bindings are really nice, this might make sense, but currently, it does not feel right to me. What do you think?

HenningHolmDE:
> Thanks! Looks very good to me. I left a few questions. I cannot test right now but will the bindings job work after a `cargo clean`? Do we expect some job order?

I just tested `cargo clean` followed by `just bindings` on my machine and it worked fine. Otherwise, I currently don't see why it would not work.

What do you mean by "job order"?

> Suggesstion for some later time: Use [`black`](https://github.com/psf/black) and maybe even [`mypy`](https://mypy-lang.org/) in CI? What do you think? :-)

Very good point! And don't forget about [Ruff](https://github.com/astral-sh/ruff). I will check how easy it is to add them to the recipe and will either add them directly or create a corresponding issue.

deny.toml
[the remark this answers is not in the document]
duesee: Oh no, let's not make it too complicated. I'm totally fine doing such things as long as it doesn't distract too much. Here, good! :-)

bindings/imap-codec-python/Cargo.toml
duesee: Good point. I agree. I think it could be separated.

deny.toml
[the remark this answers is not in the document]
HenningHolmDE: Issue #516 is now there to keep track of this.

bindings/imap-codec-python/Cargo.toml
HenningHolmDE: Switching to `0.1.0-dev1` for now. Then we can re-adress the version to use when publishing the first version to PyPI.

HenningHolmDE:
Adding checks for Black, Ruff and mypy was fairly easy, thanks for the suggestion. :+1:

I now updated the code to address all topics discussed earlier.

duesee (approved):
Excellent, thank you very much! :-)

PR #515, opened 2024-06-21
fix(build): Fix arguments to minimal version check

HenningHolmDE:
Due to the missing space at the line continuation, the arguments are directly concatenated, resulting in `imap-codec-bench--all-targets` instead of separate arguments.
For the exclude parameter, this only results in a "package ... not found" warning. However, due to the missing `--all-targets` parameter, the minimal version checks are not performed over all build targets.


duesee:
Oh well 🙈 Nice catch!

PR #517, opened 2024-06-24
Implement initial iteration of Python bindings for `imap-codec`

HenningHolmDE:
This adds the first iteration of Python bindings for `imap-codec`.

The first iteration aims to provide a first "usable" version wich a very rough API through `dict`/`tuple` types. Switching these types to provide a more Pythonic API should be achieved lateron:
- use Python classes for codec messages (e.g. `Command`, `Greeting`) -> #559
 This would make it more clear what the methods accept and return, even if these classes itself are constructed from dictionaries.
- use Python classes for `Fragment` variants (`LineFragment`, `LiteralFragment`) -> #560
 Note: This would remove the need for `Fragment` to implement `Serialize`, so it might make sense to do this before the next release of `imap-codec`.

Also, before releasing a first Python package to PyPI, these things should probably be considered:
- name of the package -> #556 
- package metadata -> #557
- some documentation on how and why to use the package -> #558


imap-codec/src/codec/encode.rs
```diff
@@ -154,6 +156,7 @@ impl Iterator for Encoded {
 }
 
 /// The intended action of a client or server.
+#[cfg_attr(feature = "serde", derive(Serialize))]
```
duesee, line 159: Ah, good catch :-)

bindings/imap-codec-python/Cargo.toml
```diff
@@ -13,4 +13,6 @@ crate-type = ["cdylib"]
 doc = false
 
 [dependencies]
+imap-codec = { version = "2.0.0-alpha.2", path = "../../imap-codec", features = ["serde"] }
```
jakoschiko, line 16: Do we want to enable all other features, too? (Similar to [`imap-next`](https://github.com/duesee/imap-next/blob/829617afe42a6987832c132f3bc52a3177dfb9d6/Cargo.toml#L22C43-L22C51))

imap-codec/Cargo.toml
```diff
@@ -19,6 +19,8 @@ default = ["quirk_rectify_numbers", "quirk_missing_text", "quirk_trailing_space"
 fuzz = []
 
 # <Forward to imap-types>
+serde = ["dep:serde", "imap-types/serde"]
```
jakoschiko, line 22: Do the CI test this feature?

jakoschiko (commented):
Looks very clean, nice!

bindings/imap-codec-python/Cargo.toml
jakoschiko, line 1: Does this crate need to be published on crates.io? If no, we could set [`publish = false`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field).

imap-codec/src/codec/encode.rs
```diff
@@ -2130,4 +2133,26 @@ mod tests {
...
+                data: b"\xCA\xFE".to_vec(),
+                mode: LiteralMode::Sync,
+            },
+            Fragment::Line {
+                data: b"\r\n".to_vec(),
+            },
+        ];
+
+        assert_eq!(
+            serde_json::to_string(&fragments).unwrap(),
```
duesee, line 2154: Thought: AFAIR, I included `serde_json` just to "show-off" some of `imap-codec`s capabilities to some Thunderbird people. But, I played with the thought to remove the `serde_json` dep in our project to spare us some compile time. Have you seen it used somewhere lese besides the `serde_json` example? Can we rewrite the test to not use `serde_json` (so that we don't start to rely on it too much?)

duesee (approved):
Thank you very much for this diligently coded PR! Let's get it in quickly!

I had one thought about `serde_json` but would not block on it.

bindings/imap-codec-python/examples/common/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+def read_more(buffer: bytearray, role: Role):
+    if not buffer:
+        prompt = "C: " if role == Role.Client else "S: "
+    else:
+        prompt = ".. "
+
+    line = read_line(prompt, role)
+
+    # If `read_line` returns `None`, standard input has been closed.
+    if line is None or line.strip() == "exit":
```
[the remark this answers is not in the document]
duesee: Nope. Should be okay.

bindings/imap-codec-python/Cargo.toml
duesee: I had the same thought but hesitated... It feels a bit off that these are features that almost everybody wants. I saw it a bit on my plate to finally make them on-by-default (inline). But... maybe we should better add them here. Otherwise there is no option to activate them.

imap-codec/src/codec/encode.rs
jakoschiko: Can we use [`serde-test`](https://crates.io/crates/serde_test)?

duesee: Argh, let's not overthink it. I'm fine with keeping this in for now :-)

jakoschiko: As always, let's create a ticket and hope that a future AI will solve it.

bindings/imap-codec-python/examples/common/__init__.py
```diff
@@ -0,0 +1,42 @@
+import sys
+from enum import Enum
+from typing import Optional
+
+COLOR_SERVER = "\x1b[34m"
+COLOR_CLIENT = "\x1b[31m"
```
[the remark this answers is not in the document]
HenningHolmDE: I like alphabetical order - done.

bindings/imap-codec-python/imap_codec.pyi
```diff
@@ -1,4 +1,156 @@
...
+    "LiteralFound" error during decoding:
+    The decoder stopped at the beginning of literal data.
+    """
+
+class Encoded:
+    """
+    An encoded message.
+
+    This struct facilitates the implementation of IMAP client- and server implementations by
+    yielding the encoding of a message through [`Fragment`]s. This is required, because the usage of
```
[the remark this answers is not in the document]
HenningHolmDE: Good point. I changed it to simple *fragments*.

bindings/imap-codec-python/src/lib.rs
```diff
@@ -1,15 +1,244 @@
...
+        };
+        Ok(encoded
+            .next()
+            .map(|value| serde_pyobject::to_pyobject(slf.py(), &value))
+            .transpose()?)
+    }
+
+    /// Dump remaining fragment data
+    fn dump(mut slf: PyRefMut<'_, Self>) -> PyResult<Bound<PyBytes>> {
+        let encoded = std::mem::take(&mut slf.0);
```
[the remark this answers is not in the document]
HenningHolmDE: Ah, thanks! I though, I'd tried `.take()` unsuccessfully before reaching for `std::mem` - mayby I'd forgotten the `.0` or something ...
Done.

bindings/imap-codec-python/imap_codec.pyi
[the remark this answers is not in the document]
HenningHolmDE: Good idea. I have now added corresponding lines to the method descriptions, e.g.:
```python
@staticmethod
def decode(bytes: bytes) -> Tuple[bytes, Tuple[()]]:
    """
    Decode idle done from given bytes.

    :param bytes: Given bytes
    :raises DecodeFailed: Decoding failed.
    :raises DecodeIncomplete: More data is needed.
    :raises DecodeLiteralFound: The decoder stopped at the beginning of literal data.
    :return: Tuple of remaining bytes and decoded idle done
    """

```

bindings/imap-codec-python/Cargo.toml
HenningHolmDE: Good idea, is now added.

bindings/imap-codec-python/src/lib.rs
```diff
@@ -1,15 +1,244 @@
...
+    }
 }
 
-/// A Python module implemented in Rust.
 #[pymodule]
 #[pyo3(name = "imap_codec")]
 fn imap_codec_python(m: &Bound<'_, PyModule>) -> PyResult<()> {
-    m.add_function(wrap_pyfunction!(sum_as_string, m)?)?;
+    m.add_class::<PyEncoded>()?;
+    m.add_class::<PyGreetingCodec>()?;
```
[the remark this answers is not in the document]
HenningHolmDE: You're welcome. When I know somethings has a deliberate order, I'm happy to keep that. Otherwise, I default to alphabetically to reduce thinking.
I now see, that I'm a bit inconsistent about where `Encoded` and the error types are sorted, but well ...

imap-codec/src/codec/encode.rs
HenningHolmDE: Replacing the dev-dep of `serde-json` with `serde-test` (or any other crate) sounds a bit unnecessary. (I have to check out `serde-test`, though.)
However, if there is a way to describe the tests without requiring an additional dependency, I would definitely go for it.

imap-codec/Cargo.toml
HenningHolmDE: It does through `--all-features`. That's why I added a tests for the additional derive, as otherwise the additional implementation would have reduced the coverage.

imap-codec/src/codec/encode.rs
HenningHolmDE: Ehm, what did I catch? 😕

bindings/imap-codec-python/Cargo.toml
HenningHolmDE: From `imap-next`:
```toml
[features]
default = ["stream"]
# ...
[dependencies]
imap-codec = { version = "2.0.0-alpha.2", features = ["starttls", "quirk_crlf_relaxed", "ext_condstore_qresync", "ext_login_referrals", "ext_mailbox_referrals", "ext_id", "ext_sort_thread", "ext_binary", "ext_metadata", "ext_uidplus"] }
```
If these are the features "almost everybody wants", it really sounds as like we should reconsider the list of default features.

It is probably a good point to think about how be able to control quirks/extensions from Python at some point. -> #547

Anyhow, which features should I enable for the Python bindings for now - exactly the list from `imap-next`?

bindings/imap-codec-python/examples/common/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+RESET = "\x1b[0m"
+
+
+class Role(Enum):
+    Client = 1
+    Server = 2
+
+
+def read_more(buffer: bytearray, role: Role):
+    if not buffer:
```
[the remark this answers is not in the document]
HenningHolmDE: If `buffer` is `None`, the caller would run into an Exception in line 28:
```
> python -c "from examples.common import read_more; read_more(None, None)"      
S: nothing

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "...\examples\common\__init__.py", line 28, in read_more
    buffer += line.encode()
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'bytes'
```

But, why stop at `None`? The parameters could actually be anything:
```
> python -c "from examples.common import read_more; read_more('hello', 'world')"
.. i like strings

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "...\examples\common\__init__.py", line 28, in read_more
    buffer += line.encode()
TypeError: can only concatenate str (not "bytes") to str
```

(Yes, the prompt is different in the two cases.)

That's all fine here, because the caller is violating the contract of the function and deserves an exception for that. Also, nothing bad has happend. (We should definetily prevent accidental misuse, like treating `None` as a wildcard to delete all entries in the database or something like that.)
Even then, the only other thing we could do here would be adding a lot of assertion code at the start of functions and then ... well ... raise an exception.

bindings/imap-codec-python/Cargo.toml
duesee: Yes. I filed https://github.com/duesee/imap-codec/issues/548

Let's not block on that here. I think it's more important to merge this PR. I can take of the features when tackling the issue.

imap-codec/src/codec/encode.rs
jakoschiko: `serde_test` has less complexity (e.g. less dependencies). IMO that's important for dev, too.

imap-codec/src/codec/encode.rs
duesee: Ah, nothing special. Just wondered if we forgot at some point to derive serde. Should we also already derive Deseralize here?

bindings/imap-codec-python/imap_codec.pyi
```diff
@@ -1,4 +1,156 @@
...
+    literals (and some other types) may change the IMAP message flow. Thus, in many cases, it is an
+    error to just "dump" a message and send it over the network.
+    """
+
+    def __iter__(self) -> "Encoded": ...
```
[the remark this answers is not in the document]
HenningHolmDE: I was assuming needing a string here, because `Encoded` is not yet defined as a type (forward-reference), which would fail at function definition time:
```
Traceback (most recent call last):
  File "...\bindings\imap-codec-python\imap_codec.pyi", line 26, in <module>
    class Encoded:
  File "...\bindings\imap-codec-python\imap_codec.pyi", line 36, in Encoded
    def __iter__(self) -> Encoded: ...
                          ^^^^^^^
NameError: name 'Encoded' is not defined
```

While this assumption was generally correct, I actually never tried running the *.pyi file through the Python interpreter. When doing so, I noticed that the order of classes was also incorrect (using `Encoded` as return value before its class definition), which I now cleaned up. As it turns out, interface stub files are usually only parsed by tools like mypy or language servers, but I like the idea that the file is actually parsable by the interpreter as well.

However, there is a different way of dealing with this as defined by [PEP 563 – Postponed Evaluation of Annotations](https://peps.python.org/pep-0563/). The idea is that annotations (e.g. type hints) are no longer evaluated during function definiton time by the interpreter which prevents the `NameError` shown above. This alternative parsing behavior is available since Python 3.7 through `from __future__ import annotations`. There do not seem to be any disadvantages in using this, so I added this as well and removed the quotes. (Actually, the reordering was unnecessary now as well, but I like the consistent order through all files.)

Thanks for sending me down this rabbit hole! 😄

bindings/imap-codec-python/Cargo.toml
HenningHolmDE: Okay, then I will leave it set to `default` + `serde` and it will extend automatically when extending `default`.

imap-codec/src/codec/encode.rs
HenningHolmDE: About `Deserialize` for `Fragment`: I did not add it intentionally because I did not need it and could not see any public interface that takes a `Fragment`, thus why would any library user want to deserialize one? Sure, we can add it for symmetry - but then we would also need to add a test to cover the implementation. :wink:

About "we forgot": Currently, nothing in `imap-codec` does implement `Deserialize`/`Serialize`. Do we want all public types to be able to implement them? Then, we should create an issue for adding these together with appropriate unit tests.

duesee: I thought about symmetry mainly. Having both pieces allows to write tests of the form `assert_eq!(x, deserialize(serialize(x))` and often comes in handy. But: we'll learn when we need it. I'll let you decide to add or not :-)

imap-codec/src/codec/encode.rs
duesee: I agree. Still, I feel that it's sometimes more important to quickly merge PRs than to get everything right. Open PRs need rebasing, CI, etc. They may block developers because a local branch already depends on the changes. Opening tickets for later™ is okay :-)

Haha, was just about to file a ticket but saw that Henning already did :-) #549

HenningHolmDE: > `serde_test` has less complexity (e.g. less dependencies). IMO that's important for dev, too.

Maybe, I'm using it not correctly yet, but currently, the reduced readibility in the tests bothers me more:

```rust
#[test]
fn test_serialize_fragment() {
let fragments = [
    Fragment::Line {
        data: b"A LOGIN alice {2}\r\n".to_vec(),
    },
    Fragment::Literal {
        data: b"\xCA\xFE".to_vec(),
        mode: LiteralMode::Sync,
    },
    Fragment::Line {
        data: b"\r\n".to_vec(),
    },
];

// using serde_json
assert_eq!(
    serde_json::to_string(&fragments).unwrap(),
    r#"[{"Line":{"data":[65,32,76,79,71,73,78,32,97,108,105,99,101,32,123,50,125,13,10]}},{"Literal":{"data":[202,254],"mode":"Sync"}},{"Line":{"data":[13,10]}}]"#
);

// using serde_test
assert_ser_tokens(
    &fragments,
    &[
        Token::Tuple { len: 3 },
        Token::StructVariant {
            name: "Fragment",
            variant: "Line",
            len: 1,
        },
        Token::Str("data"),
        Token::Seq { len: Some(19) },
        Token::U8(65),
        Token::U8(32),
        Token::U8(76),
        Token::U8(79),
        Token::U8(71),
        Token::U8(73),
        Token::U8(78),
        Token::U8(32),
        Token::U8(97),
        Token::U8(108),
        Token::U8(105),
        Token::U8(99),
        Token::U8(101),
        Token::U8(32),
        Token::U8(123),
        Token::U8(50),
        Token::U8(125),
        Token::U8(13),
        Token::U8(10),
        Token::SeqEnd,
        Token::StructVariantEnd,
        Token::StructVariant {
            name: "Fragment",
            variant: "Literal",
            len: 2,
        },
        Token::Str("data"),
        Token::Seq { len: Some(2) },
        Token::U8(202),
        Token::U8(254),
        Token::SeqEnd,
        Token::Str("mode"),
        Token::UnitVariant {
            name: "LiteralMode",
            variant: "Sync",
        },
        Token::StructVariantEnd,
        Token::StructVariant {
            name: "Fragment",
            variant: "Line",
            len: 1,
        },
        Token::Str("data"),
        Token::Seq { len: Some(2) },
        Token::U8(13),
        Token::U8(10),
        Token::SeqEnd,
        Token::StructVariantEnd,
        Token::TupleEnd,
    ],
);
```

About "less dependencies":
```
> cargo tree -p imap-codec -F serde --no-dedupe
...
[dev-dependencies]
├── serde_json v1.0.120
│   ├── itoa v1.0.11
│   ├── ryu v1.0.18
│   └── serde v1.0.204
│       └── serde_derive v1.0.204 (proc-macro)
│           ├── proc-macro2 v1.0.86
│           │   └── unicode-ident v1.0.12
│           ├── quote v1.0.36
│           │   └── proc-macro2 v1.0.86
│           │       └── unicode-ident v1.0.12
│           └── syn v2.0.70
│               ├── proc-macro2 v1.0.86
│               │   └── unicode-ident v1.0.12
│               ├── quote v1.0.36
│               │   └── proc-macro2 v1.0.86
│               │       └── unicode-ident v1.0.12
│               └── unicode-ident v1.0.12
└── serde_test v1.0.176
    └── serde v1.0.204
        └── serde_derive v1.0.204 (proc-macro)
            ├── proc-macro2 v1.0.86
            │   └── unicode-ident v1.0.12
            ├── quote v1.0.36
            │   └── proc-macro2 v1.0.86
            │       └── unicode-ident v1.0.12
            └── syn v2.0.70
                ├── proc-macro2 v1.0.86
                │   └── unicode-ident v1.0.12
                ├── quote v1.0.36
                │   └── proc-macro2 v1.0.86
                │       └── unicode-ident v1.0.12
                └── unicode-ident v1.0.12
```
Yes, we could get rid of `itoa` and `ryu`.

I'm fine going both ways but would argue for using one way for all serde tests. (Meaning updating the tests in `imap-tests` in the near future when going with `serde_test`.)

bindings/imap-codec-python/imap_codec.pyi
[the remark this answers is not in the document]
jakoschiko: Woah, thanks for the explanation. As an ousider to Python I didn't expect that this is so complicated.

bindings/imap-codec-python/examples/parse_command.py
```diff
@@ -0,0 +1,53 @@
...
+            # Parser needs more data.
+            # Read more data.
+            read_more(buffer, Role.Client)
+        except DecodeLiteralFound:
+            # Parser needs more data, and a command continuation request is expected.
+            # Simulate literal acknowledgement ...
+            print(f"S: {COLOR_SERVER}+ {RESET}")
+
+            # ... and read more data.
+            read_more(buffer, Role.Client)
```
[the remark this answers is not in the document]
jakoschiko: [This example](https://github.com/duesee/imap-codec/blob/c5ab90d7015a1abd48c0cd0c4ace789ecc3524c8/imap-codec/examples/fragmentizer.rs) mentions ctrl+d for flushing. Maybe we can do the same here.

bindings/imap-codec-python/examples/common/__init__.py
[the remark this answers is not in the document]
HenningHolmDE: Well, there is that:
```
Enter IMAP command (or "exit").

C: a SELECT {4}

S: +
.. quit 

{'tag': 'a', 'body': {'Select': {'mailbox': {'Other': {'String': {'Literal': {'data': [113, 117, 105, 116], 'mode': 'Sync'}}}}}}}
C: a SELECT {4}

S: +
.. exit

Exiting.
```

However, I would argue this is nothing we should try to prevent in the examples as these are more or less academic anyways.

bindings/imap-codec-python/examples/parse_command.py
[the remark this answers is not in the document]
HenningHolmDE: > Literals are raw bytes without line-ending. Do they need special handling when read from stdin? How do I enter them in the terminal?

You probably don't "enter" them in the terminal. However, you could send them to stdin of the process, if you are feeling fancy:
```
$ printf "a SELECT {4}\r\n\x01\x02\x03\x04\r\n" | python examples/parse_command.py
# Parsing of IMAP commands

"C:" denotes the client,
"S:" denotes the server, and
".." denotes the continuation of an (incomplete) command, e.g., due to the use of an IMAP literal.

Note: "\n" will be automatically replaced by "\r\n".

--------------------------------------------------------------------------------------------------

Enter IMAP command (or "exit").

C:
S: +
..
{'tag': 'a', 'body': {'Select': {'mailbox': {'Other': {'String': {'Literal': {'data': [1, 2, 3, 4], 'mode': 'Sync'}}}}}}}
C:
Exiting.
```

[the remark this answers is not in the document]
HenningHolmDE: > [This example](https://github.com/duesee/imap-codec/blob/c5ab90d7015a1abd48c0cd0c4ace789ecc3524c8/imap-codec/examples/fragmentizer.rs) mentions ctrl+d for flushing. Maybe we can do the same somewhere in this example.

Now, we are entering the realm of platform dependency: Simulating `EOF` through `Ctrl+D` is a *nix thing, the corresponding shortcut in Windows Terminal is `Ctrl+Z` (sometimes followed by `Enter`). Thus, the platform independent wording would be something like "EOF terminates literal".

Anyways, I now took some time to check what the current state of handling `EOF` in the current Rust and Python examples is: Neither the Rust nor the Python examples currently support sending `EOF` to end a literal.

The current "workaround" in the examples is continuing the input after the literal bytes as everything ends up in the same buffer anyways:
```
Enter IMAP command (or "exit").

C: a LOGIN {4}

S: +
.. user {4}

S: +
.. pass

{'tag': 'a', 'body': {'Login': {'username': {'String': {'Literal': {'data': [117, 115, 101, 114], 'mode': 'Sync'}}}, 'password': {'String': {'Literal': {'data': [112, 97, 115, 115], 'mode': 'Sync'}}}}}}
```

For this PR, I would argue, we should keep it like that.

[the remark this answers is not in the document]
jakoschiko: Oh, interesting. Thanks for spending time on this topic. I agree, let's keep it simple for now.

imap-codec/src/codec/encode.rs
HenningHolmDE: I now added `Deserialize` as well for symmetry to prevent any future confusion.

After implementing Python classes for `Fragment` variants (`LineFragment`, `LiteralFragment`) (which I will work on when I'm done with all of this PR's side quests :wink:), the Python bindings would (very likely) no longer require  `Fragment` to implement `Serialize` which could be removed again.

However, as removing the implementations lateron would be breaking, we should decide "Do we want all public types to to implement `Deserialize`/`Serialize` (given the `serde` feature)?" before the 2.0 release. -> #555

HenningHolmDE:
@duesee I think, I've now resolved all open questions or created issues for further tracking.

As usual, I've tried to put independent changes into separate commits for easier tracibility lateron, so you might not want to squash them.

duesee:
Thanks! Let's get this in then! 🎉

PR #527, opened 2024-06-29
build: reduce deps

duesee:


HenningHolmDE:
You can even remove `serde` and `thiserror` from the dependencies of `imap-codec` alltogether as they are currently only used in `imap-types`.

PR #530, opened 2024-07-05
refactor: Hide `bounded_static` from API

duesee:
This is a quick attempt to remove `bounded_static` from our public API. Mainly, we wrap `bounded_static::{Into,To}BoundedStatic` in `imap_types::{Into,To}Static`. This way, `imap-codec` does not need to know about `bounded_static`. Further, I feel that we almost always want to have `to_static`. Thus, I removed the feature.

Changes:

* Ungate/Remove `bounded-static` feature
* Introduce `imap_types::{ToStatic, IntoStatic}`
* Simplify `#[derive]`s

### DIFF
--- a/imap-types/src/extensions/quota.rs
+++ b/imap-types/src/extensions/quota.rs
@@ -40,8 +40,7 @@ use std::{
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::Arbitrary;
-#[cfg(feature = "bounded-static")]
-use bounded_static::ToStatic;
+use bounded_static_derive::ToStatic;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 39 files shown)


imap-types/src/extensions/quota.rs
```diff
@@ -245,9 +241,8 @@ impl<'a> QuotaGet<'a> {
 /// A type that holds a resource name and limit.
 /// Used in the SETQUOTA command.
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
-#[cfg_attr(feature = "bounded-static", derive(ToStatic))]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(ToStatic, Debug, Clone, PartialEq, Eq, Hash)]
```
jakoschiko, line 245:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq, Hash, ToStatic)]
```

jakoschiko:
I want to highlight that this is only a cosmetic change. `bounded_static` is still part of the public API. The user can see that `imap-types` is implementing traits from `bounded_static`. And the user could import and use the traits from `bounded_static` instead of the traits from `imap-types`.

However, I'm fine with this change if it's important to you.

duesee:
> I want to highlight that this is only a cosmetic change. bounded_static is still part of the public API.

Hm, you're right, unfortunately. My main goal is to become able to update `bounded-static` w/o our dependents noticing. In this regard, the changes feels like going in the right direction. But yes, it's leaky. More thoughts? Worth merging?

jakoschiko:
I don't see how we can update `bounded-static` without a breaking change. However, in practice it will be a better experience because nobody will add the dependency `bounded-static` just because of `imap-types`. So I think it's worth to merge this.

duesee:
Grr... crate organization strikes again... I would like to discuss this next monday...

PR #537, opened 2024-07-11
feat: Use `Fragmentizer` from `imap-next`

### LINKED ISSUE #536: refactor: Use `Fragmentizer` from imap-next

duesee:
Closes #536. This is only a move (with minor required changes to imports.)

### TODO

* ~Move fuzz-target~ Nevermind. Not pushed to imap-next so far.

### DIFF
--- a/imap-codec/src/fragmentizer.rs
+++ b/imap-codec/src/fragmentizer.rs
@@ -1,0 +1,1401 @@
+//! Parser for IMAP message fragments.
+
+use std::{collections::VecDeque, ops::Range};
+
+use imap_types::{
+    core::{LiteralMode, Tag},
+    secret::Secret,
+};
+
+use crate::decode::Decoder;
+
+/// Limits the size of messages that can be decoded by [`Fragmentizer`].
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub enum MaxMessageSize {
+    /// Message size is not limited.
+    ///
+    /// Using this might be dangerous because a buffer with the size of the message is
+    /// necessary. This allows an attacker to allocate an arbitrary amount of memory by
+    /// sending a large message.
+    Unlimited,
+    /// Message size is limited by the given value.
+    ///
+    /// If the size limit is exceeded then any following bytes of the current message will be
+    /// dropped. However, the fragments will still be parsed so that the end of the message can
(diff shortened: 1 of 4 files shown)


imap-codec/src/lib.rs
```diff
@@ -136,5 +137,6 @@ pub mod fuzz {
 }
 
 pub use codec::*;
+pub use fragmentizer::*;
```
jakoschiko, line 140: Are you sure about this?

- `fragmentizer` contains types with really generic names
- `fragmentizer` is only relevant for decoding

Alternatives that I like:

- Public module `imap_codec::fragmentizer`
- Public module `imap_codec::decode::fragmentizer`
- One of the options above and re-export only `imap_codec::Fragmentizer`

jakoschiko:
Should we document the `Fragmentizer` somewhere?

- `README.md`?
- `lib.rs`?

imap-codec/src/lib.rs
duesee: Let me think about it more. Didn't gave it enough thought.

duesee:
> Should we document the `Fragmentizer` somewhere?
> 
>     * `README.md`?
> 
>     * `lib.rs`?

Yes. But let's do it in a later step.

duesee:
I like the `fragmentizer` module idea. I'm not sure we already want to make it too decode-specific because I feel we may want to reuse `FragmentInfo` for encoding. I'm also a fan to not expose types twice. Thus, I think it's okay for now to have `Fragmentizer` available only via `fragmentizer::Fragmentizer` and not on the top-level, too. What do you think?

The idea of `parse_tag` was to allow users to obtain a tag on discarded data, right? So we want to expose it this way?

imap-codec/src/fragmentizer.rs
```diff
@@ -0,0 +1,1401 @@
...
+            let parsed_literal = FragmentInfo::Literal {
+                start: self.start,
+                end: self.end,
+            };
+            (parsed_byte_count, Some(parsed_literal))
+        }
+    }
+}
+
+pub fn parse_tag(message_bytes: &[u8]) -> Option<Tag> {
```
jakoschiko, line 530: I made this public by accident.
```suggestion
fn parse_tag(message_bytes: &[u8]) -> Option<Tag> {
```

PR #540, opened 2024-07-12
test(imap-codec): add `fragmentizer` fuzz-target

duesee:
Run with ...

```bash
cd imap-codec
cargo +nightly fuzz run fragmentizer --features=debug --release -- -dict=fuzz/imap.dict -max_len=128
```

(Remove `--features=debug` to suppress debug output.)

### DIFF
--- a/imap-codec/fuzz/fuzz_targets/fragmentizer.rs
+++ b/imap-codec/fuzz/fuzz_targets/fragmentizer.rs
@@ -1,0 +1,48 @@
+#![no_main]
+
+use imap_codec::fragmentizer::{Fragmentizer, MaxMessageSize};
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|input: (MaxMessageSize, Vec<&[u8]>)| {
+    let (mms, data) = input;
+
+    #[cfg(feature = "debug")]
+    println!("input: {mms:?}, {data:?}");
+
+    let mut fragmentizer = Fragmentizer::new(mms);
+
+    // Ensure `Fragmentizer` recreates input data (when smaller than `mms`).
+    let mut emitted_bytes = Vec::new();
+    let check_recreation = match mms {
+        MaxMessageSize::Unlimited => true,
+        MaxMessageSize::Limited(limit) => {
+            data.iter().map(|s| s.len()).sum::<usize>() <= limit as usize
+        }
+    };
+
+    for chunk in &data {
+        loop {
+            match fragmentizer.progress() {
+                Some(fragment_info) => {
+                    #[cfg(feature = "debug")]
+                    println!("{fragment_info:?}");
+
+                    if check_recreation {
+                        // Collect emitted bytes.
+                        emitted_bytes.extend_from_slice(fragmentizer.fragment_bytes(fragment_info));
+                    }
+                }
+                None => {
+                    fragmentizer.enqueue_bytes(chunk);
+                    break;
+                }
+            }
+        }
+    }
+
+    if check_recreation {
+        // Ensure emitted bytes are prefix of fuzzing input data.
+        let input = data.into_iter().flat_map(Vec::from).collect::<Vec<u8>>();
+        assert_eq!(input[..emitted_bytes.len()], emitted_bytes);
+    }
+});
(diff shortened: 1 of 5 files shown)


imap-codec/fuzz/fuzz_targets/fragmentizer.rs
```diff
@@ -0,0 +1,52 @@
...
+    let (mms, data) = input;
+
+    #[cfg(feature = "debug")]
+    println!("input: {mms:?}, {data:?}");
+
+    let mut fragmentizer = Fragmentizer::new(mms);
+
+    // Ensure `Fragmentizer` recreates input data (when smaller than `mms`).
+    let mut emitted_bytes = Vec::new();
+    let check_recreation = match mms {
```
duesee, line 16: Thought: Do we want to have `MaxMessageSize`? `Option<u32>` would be the same. Maybe even `u32` (with the meaning 0 is "unlimited") would be enough.

What I like about `MaxMessageSize` is that we can add good documentation. Maybe that is enough of a good reason.

jakoschiko: You can decide. I added it only for better readability.

```rust
let fragmentizer = Fragmentizer::new(MaxMessageSize::Unlimited);
```
looks better (and scarier) than
```rust
let fragmentizer = Fragmentizer::new(None);
```

Of course we also could add an `impl From<Option<u32>> MaxMessageSize`.

imap-codec/fuzz/fuzz_targets/fragmentizer.rs
```diff
@@ -0,0 +1,48 @@
...
+                None => {
+                    fragmentizer.enqueue_bytes(chunk);
+                    break;
+                }
+            }
+        }
+    }
+
+    if check_recreation {
+        // Ensure emitted bytes are prefix of fuzzing input data.
```
jakoschiko, line 44: Let's add another `println` here to see if you run into this case.

imap-codec/fuzz/fuzz_targets/fragmentizer.rs
duesee: Yeah, also thought about that. Another option would be to have a `Fragmentizer::new(u32)` and `Fragmentizer::unlimited()`. The limit could than be queried via `Fragmentizer::limit -> Option<u32>`.

duesee: Can discuss later again.

imap-codec/fuzz/fuzz_targets/fragmentizer.rs
```diff
@@ -0,0 +1,48 @@
+#![no_main]
+
+use imap_codec::fragmentizer::{Fragmentizer, MaxMessageSize};
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|input: (MaxMessageSize, Vec<&[u8]>)| {
```
jakoschiko, line 6: I think the test will look a little bit simpler if you use `Vec<Vec<u8>>` here.

imap-codec/fuzz/fuzz_targets/fragmentizer.rs
```diff
@@ -0,0 +1,48 @@
...
+                    }
+                }
+                None => {
+                    fragmentizer.enqueue_bytes(chunk);
+                    break;
+                }
+            }
+        }
+    }
+
```
jakoschiko, line 42: You could always check here `emitted_bytes <= total_bytes`.

jakoschiko (commented):
Nice test!

imap-codec/fuzz/fuzz_targets/fragmentizer.rs
jakoschiko: Please create an issue for that.

duesee:
Thanks! With the extra `emitted_bytes <= total_bytes` check, it's probably not worth it anymore to gate the `emitted_bytes` operations. I simplified.

imap-codec/fuzz/fuzz_targets/fragmentizer.rs
duesee: Already did, see #541. (Forgot to mention in my last comment, sorry.)

jakoschiko (approved):
Looks good. Don't forget to squash the commits.

Issue #535, opened 2024-07-08
Redesign `Encoded`


jakoschiko:
- Should dump take `Self`?
- Usally `Iterators` don't have other methods
- Quadratic runtime
-  ...

duesee:
* Reuse `FragmentInfo`?

PR #542, opened 2024-07-13
perf(imap-codec): Fix quadratic runtime when iterating `Encoded`

jakoschiko:
Fixes the performance issue mentioned in #535


duesee (approved):
Thanks! 🙈

PR #543, opened 2024-07-13
refactor: Remove `MaxMessageSize`

duesee:
Just something to discuss. What do you think @jakoschiko?

### DIFF
--- a/imap-codec/fuzz/fuzz_targets/fragmentizer.rs
+++ b/imap-codec/fuzz/fuzz_targets/fragmentizer.rs
@@ -1,9 +1,9 @@
 #![no_main]
 
-use imap_codec::fragmentizer::{Fragmentizer, MaxMessageSize};
+use imap_codec::fragmentizer::Fragmentizer;
 use libfuzzer_sys::fuzz_target;
 
-fuzz_target!(|input: (MaxMessageSize, Vec<Vec<u8>>)| {
+fuzz_target!(|input: (Option<u32>, Vec<Vec<u8>>)| {
     let (mms, data) = input;
 
     #[cfg(feature = "debug")]
@@ -13,7 +13,10 @@ fuzz_target!(|input: (MaxMessageSize, Vec<Vec<u8>>)| {
     let mut emitted_bytes = Vec::new();
     let data_flat = data.concat();
 
-    let mut fragmentizer = Fragmentizer::new(mms);
+    let mut fragmentizer = match mms {
+        Some(max_message_size) => Fragmentizer::new(max_message_size),
+        None => Fragmentizer::without_max_message_size(),
+    };
 
     for chunk in &data {
         loop {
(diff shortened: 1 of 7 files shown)


jakoschiko:
I don't care enough for discussing this thoroughly. If you prefer this, you can continue.

But I would prefer renaming the constructor `Fragmentizer::new_unbounded`. It's unusual to have a constructor with the name `new_*`. I would prefer `Fragmentizer::without_max_message_size`:

- It re-uses the parameter name of `Fragmentizer::new`, so it's more clear that it's the other constructor without this parameter
- It doesn't introduce another term "unbounded" that I need to understand. Does "unbounded" correlate with `max_message_size`? Or does it represent a limit in another dimension?
- The name doesn't need to be short because you should prefer `Fragmentizer::new`.

imap-codec/src/fragmentizer.rs
```diff
@@ -131,10 +111,34 @@ pub struct Fragmentizer {
 }
 
 impl Fragmentizer {
-    pub fn new(max_message_size: MaxMessageSize) -> Self {
+    /// Create `Fragmentizer`.
```
jakoschiko, line 114:
```suggestion
    /// Creates a `Fragmentizer` with maximum message size.
```

imap-codec/src/fragmentizer.rs
```diff
@@ -131,10 +111,34 @@ pub struct Fragmentizer {
...
+        Self {
+            unparsed_buffer: VecDeque::new(),
+            max_message_size: Some(max_message_size),
+            max_message_size_exceeded: false,
+            message_buffer: Vec::new(),
+            parser: Some(Parser::Line(LineParser::new(0))),
+        }
+    }
+
+    /// Create `Fragmentizer` without maximum message size.
```
jakoschiko, line 132:
```suggestion
    /// Creates `Fragmentizer` without maximum message size.
```

imap-codec/src/fragmentizer.rs
```diff
@@ -131,10 +111,34 @@ pub struct Fragmentizer {
...
+    ///
+    /// <div class="warning">
```
jakoschiko, line 134: Are we ready for this level of escalation?

imap-codec/fuzz/fuzz_targets/fragmentizer.rs
```diff
@@ -13,7 +13,10 @@ fuzz_target!(|input: (MaxMessageSize, Vec<Vec<u8>>)| {
     let mut emitted_bytes = Vec::new();
     let data_flat = data.concat();
 
-    let mut fragmentizer = Fragmentizer::new(mms);
+    let mut fragmentizer = match mms {
+        Some(max_message_size) => Fragmentizer::new(max_message_size),
+        None => Fragmentizer::without_max_message_size(),
+    };
```
jakoschiko, line 19:
```suggestion
    let mut fragmentizer =
        mms.map_or_else(Fragmentizer::without_max_message_size, Fragmentizer::new);
```

jakoschiko: Because it looks smarter.

imap-codec/src/fragmentizer.rs
duesee: `imap-codec` already has some warnings. They are yellow and cute. No red involved.

duesee:
You are lucky the CI takes so long... I already activated auto-merge because you said you don't care :D Thanks for the review :-)

jakoschiko:
I love the adrenaline when reviewing against the CI.

PR #550, opened 2024-07-14
Work on feature-gates

### LINKED ISSUE #512: question: Remove `unvalidated` feature in `imap-types`?
Currently, `imap-codec` depends on the `unvalidated` feature of `imap-types`. As the crate re-exports `imap-types`, these can be used without additional opt-in:
```rust
#!/usr/bin/env rust-script
//! Create unvalidated type
//!
//! ```cargo
//! [dependencies]
//! imap-codec = { git = "https://github.com/duesee/imap-codec.git" }
//! ```

use imap_codec::imap_types::core::Text;

fn main() {
    let text = Text::unvalidated("Hello, 🦀!");
    println!("{text:?}")
}
```
```bash
$ rust-script create_unvalidated_type.ers
Text("Hello, 🦀!")
```

Is this already known to you? For me, this is not really a large problem, as the name and the documentation of the functions do provide sufficient information on how and when to use it safely.

However, does feature gating `unvalidated` in `imap-types` still make much sense? (Without it, I would not have been suprised about the re-export.)

### LINKED ISSUE #548: build: define default features
* (Probably) enable all `quirk_*`s
* Rectify Python bindings

### LINKED ISSUE #552: build: decide what features to inline in 2.0.0

duesee:
* Closes #512
* Closes #548
* Closes #552 

Let's continue here after #517.

### DIFF
--- a/bindings/imap-codec-python/Cargo.toml
+++ b/bindings/imap-codec-python/Cargo.toml
@@ -29,5 +29,4 @@ features = [
     #"ext_login_referrals",
     #"ext_mailbox_referrals",
     "ext_metadata",
-    "ext_sort_thread",
 ]
--- a/bindings/imap-codec-python/Cargo.toml
+++ b/bindings/imap-codec-python/Cargo.toml
@@ -14,6 +14,20 @@ crate-type = ["cdylib"]
 doc = false
 
 [dependencies]
-imap-codec = { version = "2.0.0-alpha.2", path = "../../imap-codec", features = ["serde"] }
 pyo3 = "0.22.1"
 serde-pyobject = "0.4.0"
+
+[dependencies.imap-codec]
+version = "2.0.0-alpha.2"
+path = "../../imap-codec"
+features = [
+    "serde",
+    "quirk_crlf_relaxed",
+    "starttls",
+    #"ext_condstore_qresync",
+    "ext_id",
+    #"ext_login_referrals",
+    #"ext_mailbox_referrals",
+    "ext_metadata",
+    "ext_sort_thread",
+]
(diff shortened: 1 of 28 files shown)


bindings/imap-codec-python/Cargo.toml
```diff
@@ -14,6 +14,19 @@ crate-type = ["cdylib"]
...
 
 [dependencies]
-imap-codec = { version = "2.0.0-alpha.2", path = "../../imap-codec", features = ["serde"] }
 pyo3 = "0.22.1"
 serde-pyobject = "0.4.0"
+
+[dependencies.imap-codec]
+version = "2.0.0-alpha.2"
+path = "../../imap-codec"
+features = [
```
duesee, line 23: This is relevant for you, @HenningHolmDE.

This is the list of features we want (for now). `ext_condstore_qresync` is not (quite) ready. `ext_*_referrals` probably won't be ready soon (and is AFAIK a seldomly used extension that introduces quite some difficult questions.)

I think we should go with this. Maybe I manage to push the other features over the finish line soon.

duesee:
@jakoschiko, this PR mainly inlines some `ext_` feature gates. I thought about all of the features and considered the three inlined ones as stable. `unvalidated` is also not worth the effort. The other `ext_*` ones have smaller (or bigger) issues that might require a breaking API change. I would postpone that to some later™ time.

I don't think there is any review required.

bindings/imap-codec-python/Cargo.toml
HenningHolmDE: Thanks for letting me know. It probably does not change much on the level I'm currently working on.

PR #561, opened 2024-07-17
Use separate Python classes for fragment variants

### LINKED ISSUE #560: bindings(Python): Use Python classes for `Fragment` variants
Currently, the fragments of `Encoded` (Rust enum variants) are deserializes into variant dictionaries, e.g. `{"Line": {"data": list(b"a NOOP\r\n")}}`.

By using individual classes for each variant (`LineFragment`, `LiteralFragment`), handling these in the Python code would be greatly improved.

Related: This would remove the need for `Fragment` to implement `Serialize`, see #555.

HenningHolmDE:
This replaces the dictionaries returned by the `Encoded` iterator with individual `LineFragement` and `LiteralFragment` classes that can be handled more intuitively from the Python code.

Closes #560.

### DIFF
--- a/bindings/imap-codec-python/imap_codec.pyi
+++ b/bindings/imap-codec-python/imap_codec.pyi
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Tuple
+from typing import Tuple, Union
 
 class DecodeError(Exception):
     """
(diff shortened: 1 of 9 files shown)


duesee (approved):
Looks way cleaner already, thanks! (I only left a few nits.) Actually... I could just commit them and enable auto-merge :-)

Issue #555, opened 2024-07-15
feat: `Serialize`/`Deserialize` (serde) for types in `imap-codec`


HenningHolmDE:
Currently, types in `imap-codec` do not implement `Serialize`/`Deserialize`. As reuired for the Python bindings, these are now implemented for `Fragment`.

However, this requirement will probably be dropped in one of the next development iterations for the bindings. Thus, it has to be decided how to continue with `Serialize`/`Deserialize` for types in `imap-codec`. If the types shall not implement these in the long run, they should be removed for `Fragment` before the 2.0 release as this would be a breaking change.

HenningHolmDE:
@duesee Now do you want to continue on this path?

I started preparing #562, in case we do not want serialization for now.

duesee:
Let's continue with #562  :+1:

PR #562, opened 2024-07-17
Remove serialization of `Fragment`

HenningHolmDE:
Since #561, the encoded fragments are no longer deserialized into Python objects. Thus `Fragment` no longer has to implement `Serialize`.

For version 2.0, we are not aiming for implementing `Serialize`/`Deserialize` for types in `imap-codec` (see #555). Thus, removing the single implementation restores consistency for the release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1206,8 +1206,6 @@ dependencies = [
  "imap-types",
  "log",
  "nom",
- "serde",
- "serde_json",
 ]
 
 [[package]]
--- a/imap-codec/Cargo.toml
+++ b/imap-codec/Cargo.toml
@@ -18,7 +18,7 @@ default = ["quirk"]
 fuzz = []
 
 # <Forward to imap-types>
-serde = ["dep:serde", "imap-types/serde"]
+serde = ["imap-types/serde"]
 
 # IMAP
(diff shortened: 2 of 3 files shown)


HenningHolmDE:
@duesee Removing functionality that is covered by tests from something that is not at 100% line coverage means that the percentage of covered lines will fall. (Which is why coveralls is blocking the PR right now.)
How do we want to deal with this? Should I arbitrarily add a test to cover something that is currently not covered to keep the overall score up? There are several low-hanging fruits, so I‘m fine doing that, I just wanted to address the issue first.

duesee:
Hello! In general... let us not be chased by the CI :-) It should only help us keep up the quality. If we know why the coverage decreases that's fine. Especially when the decrease is not related to the changes.

HenningHolmDE:
> In general... let us not be chased by the CI :-) It should only help us keep up the quality. If we know why the coverage decreases that's fine. Especially when the decrease is not due to the addition of untested code. 

Alright, thanks for the clarificaton. Then feel free to review the PR when you find the time.

> By the way, the coverage job is not "required" for a merge. This is different for tests for example.

Oh, I overlooked that being too intimidated by the red status. 😆

duesee (approved):
Perfect, thanks!

PR #563, opened 2024-07-18
feat: add `TagGenerator`

duesee:
@jakoschiko 🤬

### DIFF
--- a/imap-types/Cargo.toml
+++ b/imap-types/Cargo.toml
@@ -16,6 +16,7 @@ exclude = [
 arbitrary = ["dep:arbitrary", "chrono/arbitrary", "chrono/std"]
 arbitrary_simplified = ["arbitrary"]
 serde = ["dep:serde", "chrono/serde"]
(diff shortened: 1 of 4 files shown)


jakoschiko:
@duesee How do we handle transitive features? We do re-export `imap-types` in `imap-codec` and `imap-next`. Do we also want to delegate features? Otherwise the benefit of re-exporting `imap-types` (you don't need to specify its version) is lost.

duesee:
Yes, I know :-/ It remembered me this was one of the reasons I thought about removing the re-export. I was too lazy to think more closely how to handle this. Do we want to simply forward all features? Only some? How do we handle name collissions?

jakoschiko:
If we

- re-export `imap-types`
- forward all features
- and suggest not depending directly on `imap-types`

then this

> Also... did you know we can break the build? By using an ext_ feature only in imap-types and not imap-codec?

should be no problem in practice.

duesee:
Yes. I think this is as close as we can get.

PR #565, opened 2024-07-19
build: Keep target dir when running coverage tests

HenningHolmDE:
Currently, running ˋjust coverageˋ clears the target directory. I introduced this as a fix in #505, however, when run locally, repeatedly clearing the entire target directory ist pretty annoying.

This changes the target directory for running the coverage measurement to ˋtarget/coverageˋ where it now lives separately from the normal build artifacts.


duesee:
Seems good to me!

(I would want to better understand later what is actually happening with the artifacts but this PR seems like a pure improvement we should merge quickly.)

PR #567, opened 2024-07-20
build: Cleanly forward features to `imap-types`

### LINKED ISSUE #566: build: cleanly forward features to `imap-types`

duesee:
Closes #566.


imap-codec/Cargo.toml
```diff
@@ -14,23 +14,10 @@ exclude = [
 
 [features]
 default = ["quirk"]
+
 # Expose internal parsers for fuzzing
 fuzz = []
 
-# <Forward to imap-types>
```
duesee, line 20: Sorry. I moved this below.

imap-codec/Cargo.toml
```diff
@@ -41,27 +28,32 @@ quirk = [
     "quirk_rectify_numbers",
     "quirk_trailing_space"
 ]
-
 # Make `\r` in `\r\n` optional.
 quirk_crlf_relaxed = []
 # Encode `A ID ()` as `A ID NIL`
-# Observed in ...
-# * GMX
```
duesee, line 49: Cleaned up a bit: We document these in modern-email/defects. We could reference it but for now I just removed.

imap-codec/Cargo.toml
```diff
@@ -41,27 +28,32 @@ quirk = [
...
 # Accept a trailing space in `STATUS` data response.
-# Observed in ...
-# * Microsoft Exchange `* STATUS INBOX (MESSAGES 100 UNSEEN 0) \n\n`
 quirk_trailing_space = []
 
+# <Forward to imap-types>
+arbitrary = ["imap-types/arbitrary"]
+arbitrary_simplified = ["imap-types/arbitrary_simplified"]
+serde = ["imap-types/serde"]
+tag-generator = ["imap-types/tag-generator"]
```
duesee, line 46: Here, `arbitrary`, `arbitrary_simplified`, and `tag_generator` are new.

imap-codec/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "imap-codec"
 description = "Rock-solid and complete codec for IMAP"
 keywords = ["email", "imap", "codec", "parser"]
 categories = ["email", "parser-implementations", "encoding", "network-programming"]
-version = "2.0.0-alpha.3"
+version = "2.0.0-alpha.4"
```
duesee, line 6: interestingly, we don't need to publish `imap-types 2.0.0-alpha.4`, too. We could let the version go out-of-sync... Which is... good?

jakoschiko: I think that's okay. What do we do the next time when we need to publish both crates? Bump both crates to the same version?

jakoschiko (approved):

imap-codec/Cargo.toml
duesee: Let's discuss next time :D Not sure... We could also just let it be out-of-sync.

PR #568, opened 2024-07-21
feat: Use separate Python classes for messages

### LINKED ISSUE #559: binding(Python): Use Python classes for codec messages
Currently, the codec messages are typeless dictionaries, e.g. `{"tag": "a", "body": "Noop"}`.

Using classes (e.g. `Command`, `Greeting`) for each of these messages would allow making it more clear what type the methods accept and return, even if these classes itself are constructed from dictionaries, e.g. `Command({"tag": "a", "body": "Noop"})`.

HenningHolmDE:
This adds wrapper classes for the messages handled by the codecs' `decode` and `encode` functions.
Currently, these classes only provide a dictionary interface through their `from_dict` and `as_dict` methods while this allows for future transitions to a more pythonic interface.

This closes #559.


bindings/imap-codec-python/imap_codec.pyi
```diff
@@ -125,21 +149,42 @@ class GreetingCodec:
         """
 
     @staticmethod
-    def encode(greeting: dict) -> Encoded:
+    def encode(greeting: Greeting) -> Encoded:
         """
         Encode greeting into fragments.
 
-        :param bytes: Given greeting
+        :param greeting: Given greeting
```
duesee, line 156: Is there a way to automatically detect typos for `:param`? (Before it was `dict` vs `bytes`.)

duesee (approved):
Thanks, Henning!

bindings/imap-codec-python/imap_codec.pyi
duesee: I created https://github.com/duesee/imap-codec/issues/572

PR #574, opened 2024-07-26
build: Remove CI path filters

HenningHolmDE:
As some CI jobs are required for PRs to be merged, these jobs have to be run even if only non-source file would be modified. Thus, the path filters for PR runs have to be removed.
Additionally, correctly managing the path lists turns out to be very error prone. Thus, removing them for the `main` branch seems to make more sense than keeping and managing them there.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -2,18 +2,8 @@ name: audit
 on:
   push:
     branches: [ main ]
-    paths:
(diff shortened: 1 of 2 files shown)


duesee:
Thanks! This bit us before and given that we want to check, e.g., PRs that change the README (because it's tested), it makes sense to just remove this filters now :+1:

PR #576, opened 2024-07-26
Extend Python bindings CI workflow

HenningHolmDE:
This adds multi-platform/version CI build and test for the Python bindings to allow for corresponding wheel releases. The GitHub workflow is based on the CI template provided by maturin.

As multi-platform/version builds are now handled by CI, there is no longer an advantage in building the bindings for multiple Python versions locally when linting/building/testing through `just bindings`. Thus, building for all detected interpreter versions is removed for the recipe.


.github/workflows/bindings_python.yml
```diff
@@ -0,0 +1,273 @@
...
+
+      - name: Download build artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: target
+
+      - name: Publish to PyPI
+        uses: PyO3/maturin-action@v1
+        env:
+          MATURIN_PYPI_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
```
duesee, line 270: I've created a [pypi.org account](https://pypi.org/user/duesee/), configured 2FA, and created the `PYPI_API_TOKEN` repository secret.

.github/workflows/bindings_python.yml
```diff
@@ -0,0 +1,273 @@
...
+      - name: Upload sdist
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheels-sdist
+          path: target/wheels
+
+  release:
+    name: Release package to PyPI
+    runs-on: ubuntu-latest
+    if: ${{ startsWith(github.ref, 'refs/tags/imap-codec-python/v') }}
```
duesee, line 242: Okay, so tags are `imap-codec-python/vx.y.z` :+1:

We could also remove the `${{ ` and ` }}` to reduce the diff from the maturin template.

.github/workflows/bindings_python.yml
```diff
@@ -0,0 +1,273 @@
...
+    needs: [linux, musllinux, windows, macos, sdist]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Assert release version matches package version
+        run: |
+          set -e
+
+          # Get release version from Git tag
+          tag_version=${GITHUB_REF#refs/tags/imap-codec-python/v}
```
duesee, line 251: Oh, nice trick! This matches `x.y.z`, right?

duesee (changes requested):
Thank you very much! Also for the great comment how to compare with the default maturin definition.

Running ...

```sh
maturin --version # maturin 1.7.0
maturin generate-ci --pytest -o ../../.github/workflows/bindings_python.yml github
git diff -R
```

... is very helpful and shows the intention of the PR very well.

We need the following changes ...

* configure branches
* configure tags
* added safety guard for versioning :+1:

The diff looks bigger because we also have non-essential changes. Notably whitespace, `.venv` vs `venv`, and `dist` vs `../../target/wheels`. Would you mind trying to reduce these changes? At least the non-essential whitespace changes? `.venv` vs `venv` could be easy as well. For the others, I would let you decide if you think it's worth it.

We can probably stay very close to the maturin template.

.github/workflows/bindings_python.yml
HenningHolmDE: Great, then we can try releasing the first development version, when this PR is completed.

.github/workflows/bindings_python.yml
HenningHolmDE: The VS Code extension for GitHub Actions did not like that syntax, thus, I thought this is deprecated syntax.

However, that seems to be a bug in the extension, as the [documentation](https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions) states:
> Note: The exception to this rule is when you are using expressions in an `if` clause, where, optionally, you can usually omit `${{` and `}}`. 

I'll go back to using double quotes then.

.github/workflows/bindings_python.yml
HenningHolmDE: Exactly, the shell parameter expansion removes the given pattern from the beginning of the value.

HenningHolmDE:
> The diff looks bigger because we also have non-essential changes. Notably whitespace, `.venv` vs `venv`, and `dist` vs `../../target/wheels`. Would you mind trying to reduce these changes? At least the non-essential whitespace changes? `.venv` vs `venv` could be easy as well. For the others, I would let you decide if you think it's worth it.

You are probably right that keeping the file as close to the template as possible makes sense. Thus, I now changed .venv/venv and dist/target back to how things are handled in the template. Now, the remaining changes should only be due to keeping the project in a subdirectory.

However, I'm not sure which whitespace changes you are referring to. I now changed the formatting of the header back to that of the template - were you referring to those?

duesee:
> However, I'm not sure which whitespace changes you are referring to. I now changed the formatting of the header back to that of the template - were you referring to those?

Weird... I was referring to this:

image

This is the output of [difftastic](https://github.com/Wilfred/difftastic). But... I have difficulties finding out where this is coming from... I suspected that it's `\n` vs `\r\n`. But maybe it's a bug in difftastic. Sorry for the distraction.

duesee (approved):
Thank you very much!

PR #579, opened 2024-07-29
chore: Add readme for Python bindings

### LINKED ISSUE #558: bindings(Python): Add basic package readme
We probably should add a basic package readme that will be display on the package's description page on PyPI ([example](https://pypi.org/project/numpy/))

The readme file is set through the `readme` key in `pyproject.toml`: https://packaging.python.org/en/latest/specifications/pyproject-toml/#readme

HenningHolmDE:
This adds a readme file for the Python bindings that should provide a basic overview of the library on the PyPI project page.

This closes #558.

Feel free to comment on points that I forgot to add and which you see as important for the first release.

### DIFF
--- a/bindings/imap-codec-python/README.md
+++ b/bindings/imap-codec-python/README.md
@@ -1,0 +1,40 @@
+# imap-codec
+
+This library provides parsing and serialization for [IMAP4rev1].
+It is based on [`imap-codec`], a building block for IMAP client and server implementations written
+in Rust, which implements the complete [formal syntax] of IMAP4rev1 and several IMAP extensions.
+
+## Usage
+
+```python
+from imap_codec import Greeting, GreetingCodec
+
+buffer = b"* OK Hello, World!\r\n<remaining>"
+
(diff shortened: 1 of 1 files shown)


bindings/imap-codec-python/README.md
```diff
@@ -0,0 +1,40 @@
+# imap-codec
+
+This library provides parsing and serialization for [IMAP4rev1] implementations.
```
duesee, line 3:
```suggestion
This library provides parsing and serialization for [IMAP4rev1].
```

bindings/imap-codec-python/examples/readme_example.py
duesee, line 1: These two files are expected to be the same, right?

I assume this won't last for long... :-) The example code in the README will -- at some point -- get out-of-sync (and maybe not even work anymore). Maybe we can run the README code in a test?

If we want to make quick progress, we can also file an issue for this and tackle later.

duesee (approved):
Thanks! Reads very good!

PR #580, opened 2024-07-29
build: Add metadata for Python bindings

### LINKED ISSUE #557: bindings(Python): Add package metadata
Before publishing to PyPI, the [package metadata](https://github.com/duesee/imap-codec/blob/main/bindings/imap-codec-python/pyproject.toml) for the Python package has to be reviewed and extended. Currently, this is mostly the content of the Maturin template.

Further information:
https://www.maturin.rs/metadata.html
https://packaging.python.org/en/latest/specifications/pyproject-toml/#declaring-project-metadata-the-project-table

HenningHolmDE:
This populates several metadata fields for the Python bindings that will be used by PyPI for the package.
As maturin merges metadata from `Cargo.toml` and `pyproject.toml` into the package (with higher precedence on `pyproject.toml`), compatible metadata was added to `Cargo.toml`.

This closes #557.

### DIFF
--- a/bindings/imap-codec-python/Cargo.toml
+++ b/bindings/imap-codec-python/Cargo.toml
@@ -1,7 +1,12 @@
 [package]
 name = "imap-codec-python"
-description = "Python bindings for imap-codec"
+description = "Python bindings for imap-codec Rust crate"
+keywords = ["email", "imap", "codec", "parser"]
 version = "0.1.0-dev1"
+authors = [
+    "Damian Poddebniak <poddebniak@mailbox.org>",
+    "Henning Holm <git@henningholm.de>",
+]
 license = "MIT OR Apache-2.0"
 edition = "2021"
 publish = false
--- a/bindings/imap-codec-python/pyproject.toml
+++ b/bindings/imap-codec-python/pyproject.toml
@@ -6,10 +6,24 @@ build-backend = "maturin"
 name = "imap-codec"
 requires-python = ">=3.8"
 classifiers = [
+    "License :: OSI Approved :: Apache Software License",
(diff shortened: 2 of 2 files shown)


bindings/imap-codec-python/Cargo.toml
```diff
@@ -1,7 +1,12 @@
 [package]
 name = "imap-codec-python"
-description = "Python bindings for imap-codec"
+description = "Python bindings for imap-codec Rust crate"
+keywords = ["email", "imap", "codec", "parser"]
```
duesee, line 4: This ...

```toml
keywords = ["email", "imap", "codec", "parser"]
```

... is also in `imap-codec/Cargo.toml`. Maybe I misunderstand, did you put this here intentionally?

bindings/imap-codec-python/Cargo.toml
```diff
@@ -1,7 +1,12 @@
...
 version = "0.1.0-dev1"
+authors = [
+    "Damian Poddebniak <poddebniak@mailbox.org>",
+    "Henning Holm <git@henningholm.de>",
+]
 license = "MIT OR Apache-2.0"
```
duesee, line 10: Theoretically... this is also already defined?

bindings/imap-codec-python/Cargo.toml
HenningHolmDE: As this is the bindings package for `imap-codec`, I thought it makes sense to use the same keywords for the Python package.

duesee (approved):
Perfect, thanks! (I asked two questions but realized they were dumb -- deleted, nevermind :-))

PR #581, opened 2024-07-29
Add CI workflow for publishing Rust crates

HenningHolmDE:
This adds a basic CI workflow for automatically publishing the Rust crates (currently `imap-codec` and `imap-types`) to crates.io when a corresponding Git tag is pushed.
For now, only the tag's version number is checked against the version number specified in the crates `Cargo.toml` file before publishing the crate is performed.

Later on, the release workflow could be further automated using crated like [`cargo-workspaces`](https://github.com/pksunkara/cargo-workspaces), [`cargo-release`](https://github.com/crate-ci/cargo-release) or [`cargo-smart-release`](https://github.com/Byron/cargo-smart-release). However, for that, a formal release workflow (when to release, which version number to bump which way and so on) has to be decided on first.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,42 @@
+name: release
+
+on:
+  push:
+    tags:
+      - 'imap-codec/v*'
+      - 'imap-types/v*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Extract crate name from Git tag
+        run: |
+          set -euo pipefail
+          tag_name=${GITHUB_REF#refs/tags/}
+          crate_name=${tag_name%/v*}
+          echo "Extracted crate name: $crate_name"
+          echo "CRATE_NAME=$crate_name" >> "$GITHUB_ENV"
+
+      - uses: actions/checkout@v4
+
+      - name: Assert release version matches crate version
+        run: |
+          set -euo pipefail
+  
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+          tag_name=${GITHUB_REF#refs/tags/}
+          crate_name=${tag_name%/v*}
+          echo "Extracted crate name: $crate_name"
+          echo "CRATE_NAME=$crate_name" >> "$GITHUB_ENV"
+
+      - uses: actions/checkout@v4
+
+      - name: Assert release version matches crate version
+        run: |
+          set -e
```
duesee, line 24:
```suggestion
          set -euo pipefail
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+      - 'imap-codec/v*'
+      - 'imap-types/v*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Extract crate name from Git tag
+        run: |
+          tag_name=${GITHUB_REF#refs/tags/}
```
duesee, line 15:
```suggestion
          set -euo pipefail
          tag_name=${GITHUB_REF#refs/tags/}
```

duesee (changes requested):

duesee (approved):

PR #592, opened 2024-08-19
feat(imap-codec): Allow `Fragmentizer` to poison messages

jakoschiko:


imap-codec/src/fragmentizer.rs
```diff
@@ -199,19 +205,35 @@ impl Fragmentizer {
         self.max_message_size_exceeded
     }
 
+    /// Returns whether the current message was explicitly poisoned to prevent decoding.
+    pub fn is_message_poisoned(&self) -> bool {
+        self.max_message_size_exceeded
```
HenningHolmDE, line 210: This should probably be `self.message_poisoned`?

jakoschiko: Yep. No tests yet.

imap-codec/src/fragmentizer.rs
```diff
@@ -199,19 +205,36 @@ impl Fragmentizer {
...
         self.message_buffer.clear();
         self.parser = Some(Parser::Line(LineParser::new(0)));
     }
 
+    /// Poisons the current message to prevent its decoding.
+    ///
+    /// When this function is called the fragments of the current message are parsed normally, but
+    /// [`Fragmentizer::decode_message`] is guaranteed to fail and return
+    /// [`DecodeMessageError::MessagePoisoned`]. This allows to skip malformed messages (e.g.
+    /// a message with a unexpected line ending) safely without the risk of treating untrusted
```
duesee, line 232:
```suggestion
    /// a message with an unexpected line ending) safely without the risk of treating untrusted
```

imap-codec/src/fragmentizer.rs
```diff
@@ -563,6 +592,8 @@ pub enum DecodeMessageError<'a, C: Decoder> {
     },
     /// Max message size was exceeded and bytes were dropped.
     MessageTooLong { initial: Secret<&'a [u8]> },
+    /// The message was explicitly poisoned to prevent decoding.
+    MessagePoisoned { undecoded: Secret<&'a [u8]> },
```
duesee, line 596: `undecoded` sounds a bit weird. `raw`? `bytes`? `message_bytes`?

duesee (approved):
Thanks!

imap-codec/src/fragmentizer.rs
jakoschiko: The other variants have descriptive names for the bytes. I like `discarded`.

PR #593, opened 2024-08-21
refactor: Move Python bindings to separate repo

HenningHolmDE:
This allows the amount of Python code of `imap-codec-python` to grow without creating a maintenance burden for non-Python developers contributing to `imap-codec`.

The code can now be found here: https://github.com/duesee/imap-codec-python


duesee:
https://github.com/duesee/imap-codec-python :-)

PR #615, opened 2024-11-12
chore: Fix failing CI

duesee:
It's theoretically not required to update the MSRV (because we don't really use `clap`). But we probably want to increase a bit due to `Arbitrary`. As long as noone is asking for it, I took the quick path and just bumped to 1.74.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -13,5 +13,8 @@ allow = [
     "Unicode-DFS-2016",
 ]
 
+[advisories]
+ignore = ["RUSTSEC-2024-0379"]
+
 [licenses.private]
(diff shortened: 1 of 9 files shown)


deny.toml
```diff
@@ -13,5 +13,8 @@ allow = [
     "Unicode-DFS-2016",
 ]
 
+[advisories]
+ignore = ["RUSTSEC-2024-0379"]
```
duesee, line 17: This is due to Stalwart. (We probably want to handle benchmarks separately.)

jakoschiko:
Let's work on #616 after this PR.

PR #625, opened 2025-01-06
imap-types: small quality of life API improvements

epilys:
I noticed some really minor pain points while using the API some days ago, so here are some fixes for it.

The major pain points would take much more time to address, unfortunately.

### DIFF
--- a/imap-types/src/core.rs
+++ b/imap-types/src/core.rs
@@ -872,6 +872,9 @@ pub struct NString<'a>(
 );
 
 impl<'a> NString<'a> {
+    /// Convenience alias to `NString(None)` for any lifetime.
+    pub const NIL: NString<'a> = NString(None);
+
     pub fn into_option(self) -> Option<Cow<'a, [u8]>> {
         self.0.map(|inner| inner.into_inner())
     }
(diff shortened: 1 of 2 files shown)


epilys:
CI failures don't seem related (too many to check them all), I made a PR to address them separately first:  #626

imap-types/src/core.rs
```diff
@@ -872,11 +872,21 @@ pub struct NString<'a>(
 );
 
 impl<'a> NString<'a> {
+    /// Convenience alias to `NString(None)` for any lifetime.
```
duesee, line 875: :+1:

duesee (commented):
I assume you intend to rebase (and mark as final) after #626. I was still curious and did a preliminary review :-) Looks good!

Feel free to already file some of the more major pain points as long as it's fresh in your head. On the good side, I feel very confident doing anything with imap-types because I started to trust that it's correct. But the cost is a bigger API surface and some annoying constructions -- happy to discuss :-)

epilys:
> On the good side, I feel very confident doing anything with imap-types because I started to trust that it's correct. But the cost is a bigger API surface and some annoying constructions -- happy to discuss :-)

I was thinking of adding some builder types, so the changes would be additive-on top of current API. Building a [`BodyStructure`](https://docs.rs/imap-types/1.0.0/imap_types/body/enum.BodyStructure.html) for example might be easier with chaining methods than writing out all the fields at once.

duesee (approved):

PR #639, opened 2025-02-07
Add quirk for missing space in continue req command

### LINKED ISSUE #638: feat: Implement quirk for missing space in command continuation response
https://github.com/modern-email/defects/issues/7

soywod:
Fixes #638.

### DIFF
--- a/imap-codec/Cargo.toml
+++ b/imap-codec/Cargo.toml
@@ -28,6 +28,7 @@ quirk = [
     "quirk_rectify_numbers",
     "quirk_trailing_space",
     "quirk_spaces_between_addresses",
+    "quirk_continue_req_space_relaxed",
     "quirk_body_fld_enc_nil_to_empty",
 ]
 # Make `\r` in `\r\n` optional.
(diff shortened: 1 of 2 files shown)


duesee:
Hey, thanks for the PR! Somehow missed this one, sorry. Do we want to allow `+<base64>\r\n`, too? Or only an additional `+\r\n`?

soywod:
This is a very good question. I would say we only encountered `+\r\n` at the moment, so let's be simple and just add a quirk for that use case. We can still extend the scope of this quirk later on.

---

Why GitHub complains about blocked merge?

> This branch must not contain merge commits.

But this branch does only contain one simple commit…

duesee:
> Why GitHub complains about blocked merge?

Does it? I can click on Rebase and merge...

soywod:
screenshot

Strange… go ahead then if it's green from your side :ok_hand:

duesee:
Weird... but do you want to adapt to only allow `+/r/n` then? If my reading is correct, we will allow `+<stuff>\r\n` after the merge? Because you said you would also prefer to have minimal quirks.

soywod:
Oh I see what you mean. In my mind it was more "this PR only makes the space after `+` optional", but it turns out to accept `+<any>\r\n`. Only accepting `+\r\n` would make the PR more complicated, I like the simplicity of this one, but at the same time the scope is larger and conflicts with what I said last time. What's your opinion?

soywod:
I opted for the smaller scope. Ready for another review!

duesee (approved):
Less code and tighter! Awesome, thanks!

PR #652, opened 2025-05-03
build: fix flake.nix

duesee:
```sh
$ nix develop
```

... lead to ...

```
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'nix-shell'
         whose name attribute is located at /nix/store/p648p3av68k4h6li0126ddkznmr1mj2s-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'buildInputs' of derivation 'nix-shell'
         at /nix/store/p648p3av68k4h6li0126ddkznmr1mj2s-source/pkgs/stdenv/generic/make-derivation.nix:395:7:
          394|       depsHostHost                = elemAt (elemAt dependencies 1) 0;
          395|       buildInputs                 = elemAt (elemAt dependencies 1) 1;
             |       ^
          396|       depsTargetTarget            = elemAt (elemAt dependencies 2) 0;

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: hash mismatch in fixed-output derivation '/nix/store/p2agx9dsvcdgfniwnanb8chhjnn2hhv9-channel-rust-stable.toml.drv':
         specified: sha256-opUgs6ckUQCyDxcB9Wy51pqhd0MPGHUVbwRKKPGiwZU=
            got:    sha256-X/4ZBHO3iW0fOenQ3foEvscgAPJYl2abspaBThDOukI=
```

This PR (seemingly) resolves the issue. However... from my understanding, this should not have been required? Did this ever work? Does the hash have something to do with `rust-toolchain.toml`? Still learning Nix... Any advice, @soywod?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,7 +36,7 @@
           # `rust-toolchain.toml` configuration file
           rust-toolchain = fenix.packages.${system}.fromToolchainFile {
             file = ./rust-toolchain.toml;
-            sha256 = "opUgs6ckUQCyDxcB9Wy51pqhd0MPGHUVbwRKKPGiwZU=";
+            sha256 = "X/4ZBHO3iW0fOenQ3foEvscgAPJYl2abspaBThDOukI=";
           };
 
         in


soywod:
The actual Nix code takes the `rust-toolchain.toml` as reference to use the right Rust toolchain. It actually points to `stable`, so every time the flake file changes the toolchain may not be the same (higher `stabel` version), therefore the hash changes as well. A solution would be to pin point to a specific version, but that would make any rustup user download that specific version rather than using his own `stable` already installed locally.

For Pimalaya projects I stopped using `rust-toolchain.toml` file as a reference, and I just take the default `stable` one from the [monthly](https://github.com/pimalaya/stream/blob/1dc5478e622dd332f2bec0ecb495ad2e22374319/flake.nix#L7) channel. I can propose a PR that would reflect those changes if you are interested in?

PR #690, opened 2026-01-21
refactor: replace warn! by debug!

soywod:
I propose to replace `warn!` by `debug!` after few discussions I got with Himalaya users. A warn is often seen as something important, sometimes even invasive, and that can be solved by users (for example, a missing configuration or a insecure setting). In `imap-codec` context, there is nothing that a lib consumer can do about a rectified malformed message. At this stage it could even be a `trace!`. Curious to know your opinion.


duesee:
Ah, yes. i think this is the same as specifying `RUST_LOG=module=trace`, right? I would have suspected that you already use something like this because otherwise you should also see quite some lower-level logs, e.g., from TLS libs?

PR #701, opened 2026-03-25
fix(imap-codec): Mail.ru FETCH interop (quoted empty, NIL body-fld-param)

### LINKED ISSUE #700: Interoperability: Mail.ru FETCH parsing failures (Himalaya / real IMAP)
## Summary

Mail.ru (and clients built on **imap-codec**, e.g. Himalaya via **imap-next** / **email-lib**) can fail to parse `FETCH` responses: the codec reports parse errors and downstream may skip entire messages, which shows up as an empty or incomplete envelope list even though the server has mail.

## What we observed

1. **Empty quoted string in addresses** — Envelope address lists use `("" NIL "user" "domain")` for an empty display name. RFC 3501 allows `quoted` to be empty (`DQUOTE DQUOTE`), but `imap-codec`’s `quoted()` parser did not accept zero characters between quotes (failure inside `nom::escaped` / `take_while1`).

2. **`body-fld-param` with `NIL` value** — Extension lists such as `("boundary" NIL)` appear in `BODYSTRUCTURE`. Strict RFC expects `string SP string` pairs; some servers send `NIL` for the value. Parsing fails without a quirk (similar spirit to existing `quirk_body_fld_enc_nil_to_empty`).

3. **Fixture / nesting** — A real line uses **`multipart/mixed`** whose first part is **`multipart/alternative`**, so the `BODYSTRUCTURE` argument correctly begins with three `(` after the keyword. Tests should reflect that shape so the full `FETCH` line parses end-to-end.

## Proposed direction

- Fix **`quoted()`** to accept `""` (minimal, spec-aligned).
- Add a **default `quirk`** feature (e.g. `quirk_body_fld_param_nil_value`) to treat `NIL` as an empty string for the second field in `body-fld-param` pairs when inside `(...)`.
- Add **regression tests**: minimal `FETCH` line for `""` in addresses; optional integration test with a captured `BODYSTRUCTURE` line; targeted parsers for envelope vs body if useful for future bisection.

Happy to open a PR with this split (strict fix vs quirk) if that matches how you want to track server deviations.

FreakyStudio:
Closes #700.

## Changes

- **`quoted()`** — Parse empty quoted strings (`""`) per RFC 3501 `*QUOTED-CHAR`, fixing envelope addresses from Mail.ru.
- **`quirk_body_fld_param_nil_value`** (default `quirk` feature) — Allow `NIL` as the value in `body-fld-param` pairs, e.g. `("boundary" NIL)` in `BODYSTRUCTURE`.
- **Tests** — Minimal `FETCH` repro; integration test + binary fixture for a real nested `multipart/mixed` → `multipart/alternative` line; `fetch` unit tests slicing envelope vs `BODYSTRUCTURE`; `response` parses full line.

## Notes

- Ran `cargo +nightly fmt` and `cargo test` under `imap-codec/`.

Maintainers: happy to drop or shrink the binary fixture if you prefer a purely `concat!`-based line.


duesee:
Nope, sorry. Too much code that feels LLM-generated. Not going to review. Also: My motivation to spend energy on mail.ru is quite low. If you can pinpoint exactly what is the problem with reproducers and a PR that is easy to review, I'll merge it. Otherwise not going to spend time on this.

PR #717, opened 2026-07-29
`LIST-EXTENDED`, part 2: capability

soywod:
Part 1: https://github.com/duesee/imap-codec/pull/716

This part 2 just adds the list extended to the capability enum.

Refs: #350

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1296,7 +1296,7 @@ dependencies = [
 
 [[package]]
 name = "imap-codec"
-version = "2.0.0-alpha.8"
+version = "2.0.0-alpha.9"
 dependencies = [
  "abnf-core",
  "base64 0.22.1",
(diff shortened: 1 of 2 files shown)


Cargo.lock
duesee, line 1: I think the Cargo.toml slipped through :-)

duesee (approved):
Thanks! Also for the great PR presentation. I started reviewing PR 1 but was not able to finish the review. This PR looked so easy, that it could be merged very quickly :-) Can you already adapt Cargo.toml here, we merge, and I'll have a look at the remaining ones soon?

By the way: I thought about introducing a feature-gate but maybe it is not worth it for `Capability`, especially since we are about to release a breaking v2.

Cargo.lock
soywod: I pushed it because CI was complaining, I think your lock is just not up-to-date on `main`. So either we push a commit to fix, or we let the lock changes on this PR, up to you.

soywod:
I precise that all 4 PRs can be merged in any order, they are independent. Once all reviewed, adjusted and merged I will send you next PRs.

Cargo.lock
duesee: Ah, understood! Let's merge then, thanks!

PR #723, opened 2026-08-07
fix: encode MODSEQ message data item with parentheses

### LINKED ISSUE #722: MessageDataItem::ModSeq is encoded without the parentheses RFC 7162's fetch-mod-resp requires
## Summary

With `ext_condstore_qresync` enabled, `MessageDataItem::ModSeq` encodes as
`MODSEQ 624140003`, but RFC 7162's `fetch-mod-resp` production requires the
value to be parenthesized: `MODSEQ (624140003)`.

The decoder is correct — it requires the parentheses — so the encoder and the
parser disagree, and a response containing this data item does not survive a
decode → encode → decode round trip.

## The grammar

RFC 7162 ("IMAP Extensions: Quick Flag Changes Resynchronization (CONDSTORE)
and Quick Mailbox Resynchronization (QRESYNC)"), §7 "Formal Syntax":

```abnf
fetch-mod-resp      = "MODSEQ" SP "(" permsg-modsequence ")"

msg-att-dynamic     =/ fetch-mod-resp

permsg-modsequence  = mod-sequence-value

mod-sequence-value  = 1*DIGIT
```

§3.1.4.2 ("MODSEQ Message Data Item in FETCH Command") states the same shape in
prose and gives Example 13:

```
Syntax:  MODSEQ ( <permsg-modsequence> )

C: a FETCH 1:3 (MODSEQ)
S: * 1 FETCH (MODSEQ (624140003))
S: * 2 FETCH (MODSEQ (624140007))
S: * 3 FETCH (MODSEQ (624140005))
S: a OK Fetch complete
```

Note that the *command* data item name is the bare `MODSEQ` (correctly encoded
by `MessageDataItemName::ModSeq`); it is only the *response* data item that is
parenthesized. That asymmetry looks like the origin of the defect.

## Observed vs. expected

| | |
| --- | --- |
| Expected | `* 1 FETCH (MODSEQ (624140003))\r\n` |
| Observed | `* 1 FETCH (MODSEQ 624140003)\r\n` |

## Reproduction

`Cargo.toml`:

```toml
[dependencies]
imap-codec = { version = "=2.0.0-alpha.9", features = ["ext_condstore_qresync"] }
```

`src/main.rs`:

```rust
use imap_codec::{ResponseCodec, decode::Decoder, encode::Encoder};

fn main() {
    // RFC 7162 §3.1.4.2, Example 13 — a well-formed MODSEQ FETCH response.
    let wire = b"* 1 FETCH (MODSEQ (624140003))\r\n";

    // The parser accepts it.
    let (_remaining, response) = ResponseCodec::default().decode(wire).unwrap();

    // The encoder drops the parentheses.
    let encoded = ResponseCodec::default().encode(&response).dump();
    println!("encoded: {:?}", String::from_utf8_lossy(&encoded));

    // And so the crate cannot parse its own output.
    println!("re-decode ok? {}", ResponseCodec::default().decode(&encoded).is_ok());
}
```

Output:

```
encoded: "* 1 FETCH (MODSEQ 624140003)\r\n"
re-decode ok? false
```

## Where it is

Encoder (the arm that omits the parentheses) — `imap-codec/src/codec/encode.rs:1780`,
in `impl EncodeIntoContext for MessageDataItem<'_>`:

```rust
#[cfg(feature = "ext_condstore_qresync")]
Self::ModSeq(value) => write!(ctx, "MODSEQ {value}"),
```

Parser (which does require them) — `imap-codec/src/fetch.rs:151`, in
`msg_att_dynamic`:

```rust
#[cfg(feature = "ext_condstore_qresync")]
let modseq = map(
    preceded(
        tag_no_case("MODSEQ "),
        delimited(char('('), mod_sequence_value, char(')')),
    ),
    MessageDataItem::ModSeq,
);
```

A one-line fix would be `write!(ctx, "MODSEQ ({value})")`.

## Affected versions

- `imap-codec` 2.0.0-alpha.9 (the current release on crates.io), with
  `imap-types` 2.0.0-alpha.7.
- **`main` is unchanged**: the same encoder arm is byte-identical at
  `imap-codec/src/codec/encode.rs:1780` on the default branch as of
  2026-08-05, so this is not something a release would pick up.

## Impact

Any server built on `imap-codec` that answers `FETCH … (MODSEQ)`,
`FETCH … (CHANGEDSINCE …)`, or sends the unsolicited MODSEQ-carrying `FETCH`
responses §3.1.4.2 and §3.2.4 call for emits a data item strict clients may
reject. It would also affect QRESYNC's resync responses, whose `FETCH`es carry
the same item.

I could not find an existing issue or pull request covering this; happy to open
a PR with the one-line change plus a round-trip test if that is useful.

marshallr12:
Fixes #722.

## The defect

RFC 7162 §7 defines the FETCH response data item as

```abnf
fetch-mod-resp     = "MODSEQ" SP "(" permsg-modsequence ")"
msg-att-dynamic    =/ fetch-mod-resp
```

but `MessageDataItem::ModSeq` encoded as `MODSEQ 624140003` — without the parentheses. The decoder (`msg_att_dynamic`) already required them, so the two disagreed and a response carrying this data item did not survive a decode → encode → decode round trip:

```
encoded: "* 1 FETCH (MODSEQ 624140003)\r\n"
re-decode ok? false
```

Note the asymmetry that likely caused this: the *command* data item name is the bare `MODSEQ`, and only the *response* data item is parenthesized. `MessageDataItemName::ModSeq` was already correct and is unchanged here.

## The change

- `imap-codec/src/codec/encode.rs` — one line, `MODSEQ {value}` → `MODSEQ ({value})`, with a note recording the command/response asymmetry so it doesn't get "simplified" back.
- `imap-codec/src/fetch.rs` — two regression tests, both gated on `ext_condstore_qresync` and both failing before the fix:
  - `test_encode_message_data_item_modseq` — known-answer encode of `MODSEQ (624140003)`.
  - `test_kat_inverse_response_fetch_modseq` — `kat_inverse_response` over all three FETCH responses from RFC 7162 §3.1.4.2, Example 13.
- Also added the `fetch-mod-resp` production to `msg_att_dynamic`'s ABNF doc comment, which previously only documented `FLAGS`.

No API change, so no SemVer impact.

## Verification

`just` isn't installed here, so I ran the CI legs directly, all with `RUSTFLAGS`/`RUSTDOCFLAGS` = `-D warnings`:

- `cargo +nightly fmt --check`
- `cargo clippy --workspace --all-targets --exclude imap-codec-bench`, and again with `--all-features`
- `cargo test --workspace --exclude imap-types-fuzz --exclude imap-codec-fuzz --all-targets --exclude imap-codec-bench`, and again with `--all-features`
- `cargo test -p imap-codec --no-default-features --features ext_condstore_qresync` (the feature in isolation)
- `cargo doc --no-deps --document-private-items --keep-going --all-features`

All green. Reverting just the encoder line turns both new tests red, so they do fence the bug.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012LQkXHvJk9Ls8v9odBhonf

### DIFF
--- a/imap-codec/src/codec/encode.rs
+++ b/imap-codec/src/codec/encode.rs
@@ -1777,7 +1777,7 @@ impl EncodeIntoContext for MessageDataItem<'_> {
                 size.encode_ctx(ctx)
             }
             #[cfg(feature = "ext_condstore_qresync")]
-            Self::ModSeq(value) => write!(ctx, "MODSEQ {value}"),
+            Self::ModSeq(value) => write!(ctx, "MODSEQ ({value})"),
         }
     }
 }
--- a/imap-codec/src/fetch.rs
+++ b/imap-codec/src/fetch.rs
@@ -139,6 +139,14 @@ pub(crate) fn msg_att(input: &[u8]) -> IMAPResult<&[u8], Vec1<MessageDataItem>> 
 /// ```
 ///
 /// Note: MAY change for a message
+///
+/// From RFC 7162 (CONDSTORE/QRESYNC):
+///
+/// ```abnf
+/// msg-att-dynamic =/ fetch-mod-resp
+///
+/// fetch-mod-resp = "MODSEQ" SP "(" permsg-modsequence ")"
+/// ```
 pub(crate) fn msg_att_dynamic(input: &[u8]) -> IMAPResult<&[u8], MessageDataItem> {
     let flags = map(
         preceded(
--- a/imap-codec/src/extensions/condstore_qresync.rs
+++ b/imap-codec/src/extensions/condstore_qresync.rs
@@ -140,7 +140,19 @@ pub(crate) fn entry_type_req(input: &[u8]) -> IMAPResult<&[u8], EntryTypeReq> {
 
 #[cfg(test)]
 mod tests {
-    use crate::response::resp_text;
+    use std::num::NonZeroU32;
+
+    use imap_types::{
+        core::Vec1,
+        fetch::MessageDataItem,
+        response::{Data, Response},
+    };
+
+    use super::*;
+    use crate::{
+        response::resp_text,
+        testing::{kat_inverse_response, known_answer_test_encode},
+    };
 
     #[test]
     fn test_condstore_qresync_codes() {
@@ -150,5 +162,50 @@ mod tests {
                 .is_ok()
         );
         assert!(resp_text(b"[HIGHESTMODSEQ 715194045007] Highest\r\n").is_ok());
+    }
+
+    #[test]
+    fn test_encode_message_data_item_modseq() {
+        known_answer_test_encode((
+            MessageDataItem::ModSeq(NonZeroU64::try_from(624140003).unwrap()),
+            b"MODSEQ (624140003)".as_ref(),
+        ));
+    }
+
+    /// RFC 7162, Section 3.1.4.2, Example 13.
+    #[test]
+    fn test_kat_inverse_response_fetch_modseq() {
+        kat_inverse_response(&[
+            (
+                b"* 1 FETCH (MODSEQ (624140003))\r\n".as_ref(),
+                b"".as_ref(),
+                Response::Data(Data::Fetch {
+                    seq: NonZeroU32::new(1).unwrap(),
+                    items: Vec1::from(MessageDataItem::ModSeq(
+                        NonZeroU64::try_from(624140003).unwrap(),
+                    )),
+                }),
+            ),
+            (
+                b"* 2 FETCH (MODSEQ (624140007))\r\n",
+                b"",
+                Response::Data(Data::Fetch {
+                    seq: NonZeroU32::new(2).unwrap(),
+                    items: Vec1::from(MessageDataItem::ModSeq(
+                        NonZeroU64::try_from(624140007).unwrap(),
+                    )),
+                }),
+            ),
+            (
+                b"* 3 FETCH (MODSEQ (624140005))\r\n",
+                b"",
+                Response::Data(Data::Fetch {
+                    seq: NonZeroU32::new(3).unwrap(),
+                    items: Vec1::from(MessageDataItem::ModSeq(
+                        NonZeroU64::try_from(624140005).unwrap(),
+                    )),
+                }),
+            ),
+        ]);
     }
 }
--- a/imap-codec/src/extensions/condstore_qresync.rs
+++ b/imap-codec/src/extensions/condstore_qresync.rs
@@ -175,37 +175,15 @@ mod tests {
     /// RFC 7162, Section 3.1.4.2, Example 13.
     #[test]
     fn test_kat_inverse_response_fetch_modseq() {
-        kat_inverse_response(&[
-            (
-                b"* 1 FETCH (MODSEQ (624140003))\r\n".as_ref(),
-                b"".as_ref(),
-                Response::Data(Data::Fetch {
-                    seq: NonZeroU32::new(1).unwrap(),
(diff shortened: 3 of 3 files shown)


imap-codec/src/codec/encode.rs
```diff
@@ -1776,8 +1776,12 @@ impl EncodeIntoContext for MessageDataItem<'_> {
                 ctx.write_all(b"] ")?;
                 size.encode_ctx(ctx)
             }
+            // Note: RFC 7162's `fetch-mod-resp` parenthesizes the value:
```
duesee: I feel we can remove all this comments when we just put the ABNF from the RFC.

imap-codec/src/fetch.rs
```diff
@@ -351,6 +355,11 @@ pub(crate) fn header_fld_name(input: &[u8]) -> IMAPResult<&[u8], AString> {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "ext_condstore_qresync")]
```
duesee: Could we move all tests to imap-codec/src/extensions/condstore_qresync.rs?

imap-codec/src/fetch.rs
```diff
@@ -493,6 +504,54 @@ mod tests {
         }
     }
 
+    // NOTE: regression test for https://github.com/duesee/imap-codec/issues/722
+    #[cfg(feature = "ext_condstore_qresync")]
+    #[test]
+    fn test_encode_message_data_item_modseq() {
+        // RFC 7162, Section 3.1.4.2: `Syntax: MODSEQ ( <permsg-modsequence> )`.
+        // Note that only the *response* data item is parenthesized; the *command* data item
```
duesee: Can we remove this comment?

duesee (changes requested):
Thank you! I left some comments inline and also have a more general request.

I feel that most comments are redundant given that we include the ABNF definition. They are not wrong but IMAP made a few... let's say... "questionable" syntax choices so what people generally should do is to ignore all prose and stick to the ABNF anyway. So I feel there is no need to point out these curiosities.

Also, I have not decided yet how to handle AI contributions and this would be the first one (co-authored: claude). Given this PR merely adjusts a string (by adding `(` and `)`) and adds some (straightforward) tests, I would ask to "redo" the PR without AI. This way I could merge it w/o having to devide how I feel about AI contributions just yet.

duesee:
Thanks! I think we can almost merge :-) One remaining question: Can you say a bit more about the tests? Do we need 3 cases? How are they different? Maybe I am missing something.

soywod:
~~No we don't really need, I just thought it could be nice to add examples directly from RFC. It can be trimmed down.~~

duesee:
I'm confused now :D Comment in wrong PR?

soywod:
Oups, it was not my PR, somehow it looked like. I need to sleep. Sorry :smile:

marshallr12:
> Thanks! I think we can almost merge :-) One remaining question: Can you say a bit more about the tests? Do we need 3 cases? How are they different? Maybe I am missing something.

The three cases inside the round-trip test are just RFC 7162 §3.1.4.2 Example 13 kept whole. They're structurally identical and any one of them catches the regression, so they're there for readability against the spec
rather than for coverage. Given your point about sticking to the ABNF and ignoring the prose, I suspect you'd rather have one — say the word and I'll collapse them.

duesee:
I see! Thanks for the clarification. I think in this particular case I would indeed prefer to collapse them.

duesee:
Thank you very much!

duesee (approved):

PR #725, opened 2026-08-28
feat: add STATUS=SIZE support

### LINKED ISSUE #704: feat(imap-types): add StatusDataItemName::Size and StatusDataItem::Size for RFC 8438 / RFC 9051
## Problem

`StatusDataItemName` and `StatusDataItem` in `imap-types` do not have `Size` variants for the `STATUS=SIZE` extension (RFC 8438), which is also a mandatory baseline item in IMAP4rev2 (RFC 9051 §6.3.10).

Unlike `StatusDataItemName::Deleted` and `StatusDataItem::DeletedStorage` (which were added as a feature-gated extension), `Size` has no representation at all — and neither enum has an `Other` escape hatch. This means:

1. A server cannot respond to `STATUS mailbox (SIZE)` requests through the typed API.
2. The library parser rejects the `SIZE` item in incoming `STATUS` commands before the application sees them.
3. Advertising `STATUS=SIZE` in `CAPABILITY` while being unable to parse or respond to it is worse than not advertising it.

## Suggested API

Following the pattern of `ext_condstore_qresync` / `DeletedStorage`:

```rust
// In StatusDataItemName
#[cfg(feature = "ext_status_size")]
Size,

// In StatusDataItem  
#[cfg(feature = "ext_status_size")]
Size(u64),
```

The value should be `u64` to match `RFC822.SIZE` semantics (sum of per-message octet counts; can exceed `u32::MAX` for large mailboxes).

## Context

This was discovered while implementing IMAP4rev2 support. `Capability::Other` provides an escape hatch for unknown capabilities, but `StatusDataItemName` has no equivalent. We filed `#702` for `Capability::Imap4Rev2` and are tracking this as a separate gap.

RFC 9051 §6.3.10 lists `SIZE`, `DELETED`, and `DELETED-STORAGE` as mandatory `STATUS` data items for IMAP4rev2. `DELETED` and `DELETED-STORAGE` are already present; `SIZE` is the missing one.

paulnivin:
Fixes #704.

Adds RFC 8438 `STATUS=SIZE` support behind the `ext_status_size` feature gate.

### DIFF
--- a/imap-codec/Cargo.toml
+++ b/imap-codec/Cargo.toml
@@ -78,6 +78,7 @@ ext_login_referrals = ["imap-types/ext_login_referrals"]
 ext_mailbox_referrals = ["imap-types/ext_mailbox_referrals"]
 ext_metadata = ["imap-types/ext_metadata"]
 ext_namespace = ["imap-types/ext_namespace"]
+ext_status_size = ["imap-types/ext_status_size"]
 ext_utf8 = ["imap-types/ext_utf8"]
 # </Forward to imap-types>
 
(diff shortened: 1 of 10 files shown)


duesee:
Perfect PR! Thank you very much!

PR #729, opened 2026-08-30
docs: Align crate-level documentation with ext_ features

### LINKED ISSUE #727: docs: Align crate-level documentation with ext_ features
I think the crate-level documentation is now missing some ext_ features. We might add them to the list when we consolidate features for the next release.

YuuGR1337:
Fixes #727

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Let's talk on [Matrix]!
 ## Features
 
 * Complete [formal syntax] of IMAP4rev1 is implemented. Furthermore, several IMAP [extensions] are supported.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,6 +17,7 @@ Let's talk on [Matrix]!
 ## Features
 
 * Complete [formal syntax] of IMAP4rev1 is implemented. Furthermore, several IMAP [extensions] are supported.
+  Extensions are enabled via `ext_` features, e.g., `ext_condstore_qresync`, `ext_id`, `ext_login_referrals`, `ext_mailbox_referrals`, `ext_metadata`, and `ext_sort_thread`. See the [extensions] documentation for the full and up-to-date list.
```
duesee, line 20: Due to its length, this list looks like it should be complete, which is negatet at the end. I would only hint to the `ext_` prefix and refer to the crate-level docs.

What I was going after with this issue is that we double-check that wr list all features in the crate-level docs since you also included a new one in your previous PR.

PR #730, opened 2026-09-22
fix!: validate UTF-8 quoted strings

paulnivin:
Rejects NUL, CR, and LF in UTF-8 quoted strings, per RFC 9755 Section 3 and RFC 3501 Section 9.

Makes `QuotedUtf8` construction fallible and contents private, matching `Quoted`. This is a breaking API change for `ext_utf8`.


duesee:
Hey @paulnivin! Thank you very much! I'm currently short on time but will try to do a proper review as soon as I can. Not sure if it looks better for @epilys :-)

I remember talking with Arnt Gulbrandsen about UTF-8 support and he was in favor of being lax in this particular case. But I do see how CR and LF could be problematic, even for some of our routines, let's fix that.

epilys (approved):
LGTM.

duesee:
Had a look on the RFC again and figured I probably misunderstood what Arnt said. Looks good to me, too! Thanks!

